PR #2, opened 2022-01-14
Add burn feature

hskang9:
I tried to add burn feature for ERC721A contract. Currently address(0) is occupied for ownership inclusion, so I used address(1) for marking burned address.

### DIFF
--- a/ERC721A.sol
+++ b/ERC721A.sol
@@ -329,7 +329,7 @@ contract ERC721A is
   }
   
   function _safeBurn(uint256 tokenId) internal {
-    _safeBurn(tokenId);
+    _safeBurn(tokenId, "");
   }
 
   /**
--- a/ERC721A.sol
+++ b/ERC721A.sol
@@ -43,6 +43,8 @@ contract ERC721A is
   uint256 private currentIndex = 0;
 
   uint256 internal immutable maxBatchSize;
+  
+  address private burned = address(1);
 
   // Token name
   string private _name;
@@ -324,6 +326,25 @@ contract ERC721A is
    */
   function _exists(uint256 tokenId) internal view returns (bool) {
     return tokenId < currentIndex;
+  }
+  
+  function _safeBurn(uint256 tokenId) internal {
+    _safeBurn(tokenId);
+  }
+
+  /**
+   * @dev Burns a certain token from id.
+   *
+   * Emits a {Transfer} event.
+   */
+  function _safeBurn(
+    uint256 tokenId,
+    bytes memory _data
+  ) internal {
+    TokenOwnership memory prevOwnership = ownershipOf(tokenId);
+    _addressData[prevOwnership.addr].balance -= 1;
+    _ownerships[tokenId] = TokenOwnership(burned, uint64(block.timestamp));
+    emit Transfer(prevOwnership.addr, burned, tokenId);
   }
 
   function _safeMint(address to, uint256 quantity) internal {


AndreiD (approved):
lgtm

hskang9:
@chiru-labs Is there a concern for this change?

ERC721A.sol
```diff
@@ -325,6 +327,25 @@ contract ERC721A is
...
+  }
+
+  /**
+   * @dev Burns a certain token from id.
+   *
+   * Emits a {Transfer} event.
+   */
+  function _safeBurn(
+    uint256 tokenId,
+    bytes memory _data
```
santiq, line 342: is there a reason for this parameter not being used?

hskang9: I thought this was needed for receiver, but feel free to remove if it is not needed.

hskang9: I want to ask for _safeMint function as well

MrMcGoats: I think it should be kept; makes it easier for other implementations to pass/use arbitrary data, should they want to

hskang9: I agree

cygaar:
Hey, thanks for the PR. We will have a more built out version of the repo soon after which we'll need you to write some tests for this. Stay tuned

ERC721A.sol
santiq: yup, I agree too, thanks for the clarification!

MrMcGoats:
Was just looking at this again, and it looks like _safeBurn(uint256) just calls itself

ERC721A.sol
```diff
@@ -325,6 +327,25 @@ contract ERC721A is
   function _exists(uint256 tokenId) internal view returns (bool) {
     return tokenId < currentIndex;
   }
+  
+  function _safeBurn(uint256 tokenId) internal {
+    _safeBurn(tokenId);
```
AndreiD, line 332: x

hskang9: yeah _safeBurn is covered with implementation

hskang9:
> Was just looking at this again, and it looks like _safeBurn(uint256) just calls itself

no @MrMcGoats, as you can see from @AndreiD, it has another function to cover implementation like _safeMint

harry830622:
You should check `tokenId+1` and update it accordingly just like `transfer` or the tokens after `tokenId` will be burnt as well.
Also using `0x0000...dead` for burn address would be more intuitive IMO.

harry830622:
In addition, most of the view functions need to be updated to consider the burnt tokens.

hskang9:
> You should check `tokenId+1` and update it accordingly just like `transfer` or the tokens after `tokenId` will be burnt as well.
> Also using `0x0000...dead` for burn address would be more intuitive IMO.

This is true. I will add that

hskang9:
> In addition, most of the view functions need to be updated to consider the burnt tokens.

I am not sure if there is a view function to check burns in usual ERC721 though. For the backward compatibility, I will pass on this and let frontend handle.

MrMcGoats:
> > Was just looking at this again, and it looks like _safeBurn(uint256) just calls itself
> 
> no @MrMcGoats, as you can see from @AndreiD, it has another function to cover implementation like _safeMint
@hskang9 
I'm not seeing any other function defined as `_safeMint(uint256)` apart from the one on line 331. However, if that exists, I don't see the point in wrapping it with a function that has the exact same name and parameters.

As far as I can tell, this is not done with `_safeMint(address,uint256)`. The only implementation of that function that I can find (line 350) calls `_safeMint(address,uint256,bytes)`

hskang9:
@MrMcGoats I think you can see it in line 340 for `_safeBurn`'s implementation. Same as `_safeMint`, `_safeBurn` impl function has bytes data_ as a parameter just in case of receiver as referred before. Would you mind checking the whole logs of comments?

ERC721A.sol
```diff
@@ -43,6 +43,8 @@ contract ERC721A is
   uint256 private currentIndex = 0;
 
   uint256 internal immutable maxBatchSize;
+  
+  address private burned = address(1);
```
ahbanavi, line 47: What's the reason for using `address(1)` instead of `address(0)`?

ahbanavi:
I think we should close this, cause #61 is more likely to be merged soon.

cygaar:
Thanks for the contribution, closing this since we've merged in burn functionality

PR #5, opened 2022-01-17
Ensure that a mint quantity does not exceed collection size

aksperiod:
I was adding this to multiple functions when I realized I would just shim my own version of ERC721A that had this check in the most sensible place.

### DIFF
--- a/ERC721A.sol
+++ b/ERC721A.sol
@@ -350,6 +350,7 @@ contract ERC721A is
     // We know if the first token in the batch doesn't exist, the other ones don't as well, because of serial ordering.
     require(!_exists(startTokenId), "ERC721A: token already minted");
     require(quantity <= maxBatchSize, "ERC721A: quantity to mint too high");
+    require(startTokenId + quantity < collectionSize, "ERC721A: would exceed max supply");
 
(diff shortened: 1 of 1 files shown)


cygaar:
collectionSize was actually removed from ERC721A so this is no longer valid

aksperiod:
i see. any specific reason why was it removed? are there many use-cases where collections have no limit?

cygaar:
We just didn't want to cap the limit - there may be use cases that don't require a limit and we don't want to put that restriction in

aksperiod:
I think that's probably wise, but what if it was an _optional_ second parameter that was otherwise initialized as `uint256(-1)`. I am only saying this because I assume there are (many) more fixed-size ERC721A contracts than not. For now at least. Just a thought.

cygaar:
We could potentially add that in or as an extension. We'll put some thought into it

rosman83:
Shouldn't you be instead ensuring that the mint quantity in the user mint request is limited to the amount that all users can mint in the first place (assuming a standard NFT that limits each user to mint X amounts per TX)

chiru-labs:
Closing because collectionSize has been removed

PR #6, opened 2022-01-20
Create SimpleERC721A.sol

0xJacobean:
This is a super straightforward implementation of the ERC721A with the basic set of features I think a 10K project would need. Thanks!

### DIFF
--- a/SimpleERC721A.sol
+++ b/SimpleERC721A.sol
@@ -1,0 +1,95 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "./ERC721a.sol";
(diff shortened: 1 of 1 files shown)


AndreiD:
good idea. add merkle root whitelist and it's good to go ;)

cygaar:
We want to keep this repo for just ERC721A - this seems to be more of an implementation, and we don't necessarily need that.

AndreiD:
this could be in some folder like "example". I think it's a good idea for new developers to see how this library gets integrated.

AndreiD:
ah, now I see you wrote in the README a tutorial. nevermind then

PR #21, opened 2022-01-24
Implements batching of batches inside of safeMint

0xface1e555:
Modifies `safeMint` to be able to accept an arbitrary quantity which should simplify the API for a lot of use cases. 

Drawbacks: has a  minor impact on gas for the mint of 1 - `77395 -> 78188`. (optimizer enabled at 200)

For the cases of minting 10 with a maxBatchSize of 5 the overall gas is: 121709
Screen Shot 2022-01-23 at 10 45 35 PM

Your token 0 is 15k gas more expensive than other tokens because `currentIndex` is not "warm". So tested it on tokens 1-10 to get proper gas.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,6 @@
 {
+  "printWidth": 120,
+  "singleQuote": true,
   "overrides": [
     {
       "files": "*.sol",
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -317,23 +317,17 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
             addressData.numberMinted + uint128(quantity)
         );
 
-        // Recording ownerships in increments of maxBatchSize.
-        uint256 remainingTokens = quantity;
-        uint256 ownershipTokenId = startTokenId;
-
-        while (remainingTokens > 0) {
-            if (remainingTokens > maxBatchSize) {
-                _ownerships[ownershipTokenId] = TokenOwnership(to, uint64(block.timestamp));
-                ownershipTokenId += maxBatchSize;
-                remainingTokens -= maxBatchSize;
-                continue;
-            }
-
-            // Last batch
-            _ownerships[ownershipTokenId] = TokenOwnership(to, uint64(block.timestamp));
-            remainingTokens = 0;
+        uint256 batches = quantity / maxBatchSize;
+        if (quantity % maxBatchSize != 0) {
+            batches += 1;
         }
 
+        // Recording ownerships in increments of maxBatchSize.
+        for (uint256 i = 0; i < batches; i++) {
+            _ownerships[startTokenId + maxBatchSize * i] = TokenOwnership(to, uint64(block.timestamp));
+        }
+
+        // Dispatching all events sequentially
         for (uint256 i = startTokenId; i < startTokenId + quantity; i++) {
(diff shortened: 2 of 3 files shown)


.prettierrc
```diff
@@ -1,4 +1,6 @@
 {
+  "printWidth": 120,
+  "singleQuote": true,
```
0xface1e555, line 3: When I was modifying tests I couldn't figure out what settings were used when formatting, decided to add some to make it more consistent. Any suggestions on what you guys want to use?

contracts/ERC721A.sol
```diff
@@ -316,20 +316,26 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
             addressData.balance + uint128(quantity),
             addressData.numberMinted + uint128(quantity)
         );
-        _ownerships[startTokenId] = TokenOwnership(to, uint64(block.timestamp));
 
-        uint256 updatedIndex = startTokenId;
+        uint256 batches = quantity / maxBatchSize;
+        if (quantity % maxBatchSize != 0) {
+            batches += 1;
+        }
```
0xface1e555, line 323: e.g.
Minting 10 - 2 batches
Minting 13 - 3 batches

contracts/ERC721A.sol
```diff
@@ -316,20 +316,26 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
+
+        // Recording ownerships in increments of maxBatchSize.
+        for (uint256 i = 0; i < batches; i++) {
+            _ownerships[startTokenId + maxBatchSize * i] = TokenOwnership(to, uint64(block.timestamp));
+        }
```
0xface1e555, line 328: First i = 0, so we always start with `startTokenId` because `maxBatchSize * i` will zero itself out.

test/ERC721A.test.js
```diff
@@ -258,15 +254,9 @@ describe('ERC721A', function () {
       });
 
       it('rejects mints to the zero address', async function () {
-        await expect(
-          this.erc721a['safeMint(address,uint256)'](ZERO_ADDRESS, 1)
-        ).to.be.revertedWith('ERC721A: mint to the zero address');
-      });
-
-      it('rejects quantity > maxBatchSize', async function () {
```
0xface1e555, line 266: removed this test, sorry for a weird diff in this file

contracts/ERC721A.sol
```diff
@@ -316,20 +316,26 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
             addressData.balance + uint128(quantity),
             addressData.numberMinted + uint128(quantity)
         );
-        _ownerships[startTokenId] = TokenOwnership(to, uint64(block.timestamp));
 
-        uint256 updatedIndex = startTokenId;
+        uint256 batches = quantity / maxBatchSize;
```
cygaar, line 320: I think if we do go with this change, we should rename `maxBatchSize` to `batchSize` since it's not really a constraint, and more of a parameter used to optimize the mint process

fulldecent:
This implementation will  be easier with a `+= MAX_BATCH_SIZE`.

No comment yet on whether this change is actually good.

ahbanavi:
@0xface1e555, thanks for your contributions, but this should be closed since `batchSize` is removed in #28.  

CC: @cygaar @Vectorized

PR #4, opened 2022-01-17
Made Upgradeable version

MrMcGoats:
Made an upgradeable version of ERC721A

Also added hardhat files, so the contract can easily be built and deployed.

Tests by @cygaar (https://github.com/cygaar/ERC721A/tree/version2)

PR #25, opened 2022-01-24
Extend openzeppelin

MrMcGoats:
Modified contract to extend the OpenZeppelin implementation of ERC721

This branch includes an upgradeable version of ERC721A (ERC721AUpgradeable). A separate pull request for an upgradeable version that does not extend the OpenZeppelin ERC721Upgradeable implementation is #4.

I believe that extending the OpenZeppelin implementations simplifies this contract, as not every feature needs to be re-written from scratch. I also needed the legacy `_mint(address,uint256)` because I need the ability to mint items with specific item IDs for use with [Xenum-Technology/cross-chain-bridge-nfts](https://github.com/Xenum-Technology/cross-chain-bridge-nfts).


cygaar:
Hey @MrMcGoats can you try to isolate your PR to the smaller changes? Can you remove the typescript changes and only include the upgradeable changes? It's super difficult to review large PRs and ensure that every change is vetted thoroughly

MrMcGoats:
Do you mean just remove `hardhat.config.ts` ?

cygaar:
Yes and the tsconfig. We should also remove any added linters here - that should be in a separate PR that the community can decide up on

MrMcGoats:
Ok, np.

I don't think I added any linters myself, but I agree that should be a separate PR

MrMcGoats:
Oh, right. Hardhat config sets up linters. I'll remove them

cygaar:
You also added a new eslint settings file. You also linted several files which has resulted in a couple changes that I don't think you intended to be part of this PR

MrMcGoats:
I just ran `npm run lint:sol` and that seems to have corrected the mistake. Lmk if you find anything else though, of course.

Also, not sure what eslint settings file you're referring to. I don't see one in the changed file list

cygaar:
Will take a look at this soon

contracts/ERC721A.sol
```diff
@@ -428,7 +411,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
      * @param _data bytes optional data to send along with the call
      * @return bool whether the call correctly returned the expected magic value
      */
-    function _checkOnERC721Received(
+    function _checkOnERC721AReceived(
```
fulldecent, line 414: This function name could be better as staying generic since the -A version does not change anything over baseline ERC-721.

MrMcGoats: I agree, but I had to change it because openzeppelin's ERC721Upgradeable has it as a non-overrideable private function and I needed to be able to call it in ERC721A

AboldUSER:
Is this good enough to be used currently?

MrMcGoats:
Yes, but it's not up to date with the most recent ERC-721A changes, so depends exactly what features you need.

I use it in [this](https://github.com/Xenum-Technology/cross-chain-bridge-nfts) project and it does everything I need.

PR #27, opened 2022-01-25
add require quantity > 0 in _safeMint()

### LINKED ISSUE #18: Comments on ERC721A
I'd like to start off by thanking you for working on this and allowing contributions from the community.

Here are some of my comments on the current implementation and what I personally would change.

1. Remove maxBatchSize_ in constructor
 - keeps implementation as close as possible to the original
 - allows max batch logic to be implemented in inheriting class, just like before
 - as I see it, there is no need for ERC721A to know the max batch size
 
image

...

image

2. Add functionality for starting index to be either 0 or 1 (this has been discussed).
 - even though I believe they should simply start at 0, some libraries like hashlips require them to start at 1 by default. That's why it would be nice to have an easy way of switching this in the contract

3. Add _mint alongside _safeMint (and perhaps advocate this as the standard).
 - currently _safeMint checks that caller is a ERC721Receiver n times in a loop
 - this adds unnecessary checks, as the standard seems to be to disallow minting from contracts

image
- also added a require for quantity to be > 0, as otherwise someone could technically be added as an owner (at least temporarily), even though the counter doesn't increase

Other than that, I'm not sure how I feel about supporting the IEC721Enumerable interface, if the functionality is added by inefficient loops that should never be called on-chain. I don't see much point in `tokenOfOwnerByIndex` at all (it has it's use-case in ERC721Enumerable). Why not replace that with something actually useful, such as listing all tokenIds of an owner (though this could also be considered for an extension).

image

A last note on 1:
I did actually see a case being made for adding the maximum collection size into ERC721A, because otherwise the inheriting contract will need to read from the chain (get totalSupply and make sure it's below max collections size), as opposed to the mint function handling this for no extra read cost (totalSupply is already present).

I haven't submitted a pull request yet, because I would like to see some reactions to this first.

willisk:
_safeMint is missing a require quantity > 0. This can lead to odd behaviour.
Resolves #18


cygaar (changes requested):
Please add tests for this

willisk:
Added a test case. 
Like someone had mentioned before, I was struggling to get the correct prettier formatting settings. It seems like the .prettierrc only specifies the settings for .sol files (but they aren't even applied, because you have double quotes instead of single quotes)..

cygaar:
I added in your lint settings in a separate PR so you should be able to rebase and have no issues

willisk:
The prettier outputs still don't seem to match up for the .js files. Could this be due to different prettier / eslint versions?
npx eslint -v
v8.7.0
npx prettier -v
2.5.1

cygaar:
I have the same versions as you... odd. I'm ok with the lint changes that you on your end though

cygaar (approved):

PR #28, opened 2022-01-25
Remove `batchSize`

### LINKED ISSUE #18: Comments on ERC721A

willisk:
`batchSize` is not required. The implementation should be kept as close to the original as possible. Requiring a certain batch size is better dealt with in inheriting contracts.
Resolves #18

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -38,8 +38,6 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
 
     uint256 internal currentIndex = 0;
 
-    uint256 internal immutable maxBatchSize;
-
     // Token name
     string private _name;
 
@@ -59,19 +57,9 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
     // Mapping from owner to operator approvals
     mapping(address => mapping(address => bool)) private _operatorApprovals;
 
-    /**
-     * @dev
-     * `maxBatchSize` refers to how much a minter can mint at a time.
-     */
-    constructor(
-        string memory name_,
-        string memory symbol_,
-        uint256 maxBatchSize_
-    ) {
-        require(maxBatchSize_ > 0, 'ERC721A: max batch size must be nonzero');
+    constructor(string memory name_, string memory symbol_) {
         _name = name_;
         _symbol = symbol_;
-        maxBatchSize = maxBatchSize_;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


chiru-labs (commented):
Please revert styling changes to keep PR changes isolated, rest lgtm

Let's also cut a new version after this goes in because the constructor is a backcompat break

willisk:
I added those the 2 lines in the .prettierrc to because I couldn't get it to look like before with any settings. If you are using specific settings, please add them to the .prettierrc (or describe exactly how you are formatting the files). Currently it looks like only .sol files have styling settings, these however are not even being applied correctly, because the .prettierrc specifies singleQuote: true, but the .sol files currently use double quotes.
If I can figure out how to make the styling match, I'll push again.

chiru-labs:
Got it thanks. How about first pushing the prettier changes as a separate PR to keep things isolated? Can fast-follow approve that one to keep iteration quick

willisk:
Did you want me to change this to a new version? Is that done under `version` in package.json?

cygaar:
Nah I can handle the version update when we cut a new release

cygaar (approved):

PR #29, opened 2022-01-25
Remove ERC721Enumerable interface

willisk:
Removes ERC721Enumerable interface. The functionality introduced by the interface has its use in ERC721Enumerable.sol, however here it does not add any value. 
In addition, the function `tokenOfOwnerByIndex(address owner, uint256 index)` should never be called on-chain. To remove any possibility of an accidental call on-chain, I propose to remove the interface (except for totalSupply()).
See #18

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -6,7 +6,6 @@ pragma solidity ^0.8.0;
 import "@openzeppelin/contracts/token/ERC721/IERC721.sol";
 import "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 import "@openzeppelin/contracts/token/ERC721/extensions/IERC721Metadata.sol";
-import "@openzeppelin/contracts/token/ERC721/extensions/IERC721Enumerable.sol";
 import "@openzeppelin/contracts/utils/Address.sol";
 import "@openzeppelin/contracts/utils/Context.sol";
 import "@openzeppelin/contracts/utils/Strings.sol";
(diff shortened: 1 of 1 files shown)


chiru-labs:
@willisk would you be open to opening a PR that makes ERC721AEnumerable as an extension, similar to https://github.com/chiru-labs/ERC721A/pull/34, where we can document tradeoffs involved with calling onchain from a Solidity write path?

contracts/ERC721A.sol
```diff
@@ -6,7 +6,6 @@ pragma solidity ^0.8.0;
 import "@openzeppelin/contracts/token/ERC721/IERC721.sol";
 import "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 import "@openzeppelin/contracts/token/ERC721/extensions/IERC721Metadata.sol";
-import "@openzeppelin/contracts/token/ERC721/extensions/IERC721Enumerable.sol";
 import "@openzeppelin/contracts/utils/Address.sol";
 import "@openzeppelin/contracts/utils/Context.sol";
 import "@openzeppelin/contracts/utils/Strings.sol";
```
coindegen, line 11: in the comment on line 16 you can delete the "Enumerable" part of "including the Metadata and Enumerable extension".

chiru-labs:
@willisk bumping above comment- otherwise we can work off this branch to add ERC721AEnumerable

syffs:
@chiru-labs please see #81

cygaar:
@willisk we ended up merging #107 because it was based off of the latest main. Regardless, thanks for opening up this PR

PR #34, opened 2022-01-27
Move _setOwnersExplicit to separate extension 

cygaar:
_setOwnersExplicit doesn't need to be in the base level contract as most users won't need it. However, we want to provide it as an optional extension in case some users want to explicitly set ownerships after a mint is over.

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -34,7 +34,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
         uint128 numberMinted;
     }
 
-    uint256 private currentIndex = 0;
+    uint256 internal currentIndex = 0;
 
     uint256 internal immutable maxBatchSize;
 
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -46,7 +46,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
 
     // Mapping from token ID to ownership details
     // An empty struct value does not necessarily mean the token is unowned. See ownershipOf implementation for details.
-    mapping(uint256 => TokenOwnership) private _ownerships;
+    mapping(uint256 => TokenOwnership) internal _ownerships;
 
     // Mapping owner address to address data
     mapping(address => AddressData) private _addressData;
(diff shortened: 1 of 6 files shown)


contracts/ERC721A.sol
```diff
@@ -46,7 +46,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
 
     // Mapping from token ID to ownership details
     // An empty struct value does not necessarily mean the token is unowned. See ownershipOf implementation for details.
-    mapping(uint256 => TokenOwnership) private _ownerships;
+    mapping(uint256 => TokenOwnership) internal _ownerships;
```
cygaar, line 49: Had to convert this to internal in order for ERC721AOwnersExplicit to read the variable. We can also redeclare ownerships in that file, I have no strong preference here

cygaar: Same with currentIndex above

contracts/ERC721A.sol
```diff
@@ -395,29 +395,6 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
-     */
-    function _setOwnersExplicit(uint256 quantity) internal {
-        uint256 oldNextOwnerToSet = nextOwnerToExplicitlySet;
-        require(quantity > 0, "quantity must be nonzero");
-        uint256 endIndex = oldNextOwnerToSet + quantity - 1;
-        if (endIndex > currentIndex - 1) {
-            endIndex = currentIndex - 1;
-        }
-        // We know if the last one in the group exists, all in the group exist, due to serial ordering.
-        require(_exists(endIndex), "not enough minted yet for this cleanup");
```
cygaar, line 411: This line will never evaluate to true due to line 408 so I removed this check in the refactored function

contracts/extensions/ERC721AOwnersExplicit.sol
```diff
@@ -0,0 +1,34 @@
...
+abstract contract ERC721AOwnersExplicit is ERC721A {
+    uint256 public nextOwnerToExplicitlySet = 0;
+
+    /**
+     * @dev Explicitly set `owners` to eliminate loops in future calls of ownerOf().
+     */
+    function _setOwnersExplicit(uint256 quantity) internal {
+        require(quantity > 0, "quantity must be nonzero");
+        require(currentIndex > 0, "no tokens minted yet");
+        require(nextOwnerToExplicitlySet < currentIndex, "all ownerships have been set");
```
cygaar, line 17: Added this to prevent unnecessary calls to this function

chiru-labs (approved):

PR #48, opened 2022-01-30
Fix implementation example with correct number of constructor arguments

shakiXBT:
The current implementation example of ERC721A provided in the README.md file is not working.

ERC721A inherited constructor requires only 2 arguments since the argument `uint256 maxBatchSize_` has been removed. The example has yet to be updated and still passes the `maxBatchSize_` arguments, leading to a compilation error.


shakiXBT:
Didn't bother making a feature branch since this is just a one-liner in the docs.

chiru-labs (approved):
Nice catch!

PR #54, opened 2022-01-31
Assigning values to structs individually instead of a single assignment

DaniPopes:
[Cheaper by about 230 gas per mint and 15,000 gas on deployment.](https://github.com/DaniPopes/soltracts/commit/dd75437e42bd696383b5a859b06e1913055909a8) These savings are just for the mint ones, didn't test transfer.

[Source:](https://medium.com/coinmonks/gas-optimization-in-solidity-part-i-variables-9d5775e43dde)
> It is more gas efficient to initialize a tightly packed struct with separate assignments instead of a single assignment. Separate assignments makes it easier for the optimizer to update all the variables at once.

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -296,12 +296,11 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
 
         _beforeTokenTransfers(address(0), to, startTokenId, quantity);
 
-        AddressData memory addressData = _addressData[to];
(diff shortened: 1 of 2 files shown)


chiru-labs (approved):

cygaar (changes requested):
Can you make the corresponding changes to ERC721AOwnersExplicit as well?

DaniPopes:
Yep! Done.

cygaar (approved):

PR #59, opened 2022-01-31
Add gas optimizations

DaniPopes:
Wrote up a few optimizations on the main contracts, open to feedback especially for [the unchecked blocks](https://github.com/chiru-labs/ERC721A/commit/ae1cdcff0eb73c3b5a12034664a260a60486791f) if you think that over/underflow checks are needed in those functions.

[1b32797:](https://github.com/chiru-labs/ERC721A/commit/1b32797fd57660f2463500960d89bedf9f1cf8a5) _exists(uint256 id) returns id < currentIndex, and we call it with currentIndex, which always results in it returning false.

[ae1cdcf:](https://github.com/chiru-labs/ERC721A/commit/ae1cdcff0eb73c3b5a12034664a260a60486791f) Explanations in the code comments.

[5fca737:](https://github.com/chiru-labs/ERC721A/commit/5fca737705af62cf3c04f546fc9e1a588a03e846) Initializing a variable with a 0 is more expensive than not doing so, even if both results are the same. [Source:](https://medium.com/coinmonks/gas-optimization-in-solidity-part-i-variables-9d5775e43dde) 
> uint256 value; is cheaper than uint256 value = 0;

[c23e2e0:](https://github.com/chiru-labs/ERC721A/commit/c23e2e07b41453f3f366f6b0a1edb281aa6d5c3d) All values are integers which cannot be negative, saves a few opcodes.

[ee2abe2:](https://github.com/chiru-labs/ERC721A/commit/ee2abe273c77060e25535ac99876c073f4d9df40) Previous changes applied to `ERC721AOwnersExplicit`.

[81daf25:](https://github.com/chiru-labs/ERC721A/commit/81daf257a9443d24da2aa5707ef303c09c77c664) Loading `nextOwnerToExplicitlySet` into memory saves an SLOAD opcode later on line 23.

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -294,8 +294,6 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
     ) internal {
         uint256 startTokenId = currentIndex;
         require(to != address(0), 'ERC721A: mint to the zero address');
-        // We know if the first token in the batch doesn't exist, the other ones don't as well, because of serial ordering.
-        require(!_exists(startTokenId), 'ERC721A: token already minted');
         require(quantity > 0, 'ERC721A: quantity must be greater than 0');
 
         _beforeTokenTransfers(address(0), to, startTokenId, quantity);
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -36,7 +36,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
         uint128 numberMinted;
     }
 
-    uint256 internal currentIndex = 0;
+    uint256 internal currentIndex;
 
     // Token name
     string private _name;
@@ -85,12 +85,12 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
     function tokenOfOwnerByIndex(address owner, uint256 index) public view override returns (uint256) {
         require(index < balanceOf(owner), 'ERC721A: owner index out of bounds');
         uint256 numMintedSoFar = totalSupply();
-        uint256 tokenIdsIdx = 0;
-        address currOwnershipAddr = address(0);
+        uint256 tokenIdsIdx;
+        address currOwnershipAddr;
 
         // Counter overflow is incredibly unrealistic.
         unchecked {
-            for (uint256 i = 0; i < numMintedSoFar; i++) {
+            for (uint256 i; i < numMintedSoFar; i++) {
                 TokenOwnership memory ownership = _ownerships[i];
                 if (ownership.addr != address(0)) {
                     currOwnershipAddr = ownership.addr;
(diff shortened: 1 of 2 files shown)


contracts/extensions/ERC721AOwnersExplicit.sol
```diff
@@ -6,30 +6,37 @@ pragma solidity ^0.8.0;
...
-            if (_ownerships[i].addr == address(0)) {
-                TokenOwnership memory ownership = ownershipOf(i);
-                _ownerships[i].addr = ownership.addr;
-                _ownerships[i].startTimestamp = ownership.startTimestamp;
+            // Set the end index to be the last token index
+            if (endIndex + 1 > currentIndex) {
+                endIndex = currentIndex - 1;
             }
+
+            for (uint256 i = oldNextOwnerToSet; i <= endIndex; i++) {
```
chiru-labs, line 31: Seems like we don't need oldNextOwnerToSet anymore and can just use `_nextOwnerToExplicitlySet` here

contracts/extensions/ERC721AOwnersExplicit.sol
```diff
@@ -6,30 +6,37 @@ pragma solidity ^0.8.0;
...
-        if (endIndex > currentIndex - 1) {
-            endIndex = currentIndex - 1;
-        }
+        require(quantity != 0, 'quantity must be nonzero');
+        require(currentIndex != 0, 'no tokens minted yet');
+        uint256 _nextOwnerToExplicitlySet = nextOwnerToExplicitlySet;
+        require(_nextOwnerToExplicitlySet < currentIndex, 'all ownerships have been set');
+
+        // Index underflow is impossible.
+        // Counter or index overflow is incredibly unrealistic.
```
chiru-labs, line 21: Please elaborate a bit more on this line (e.g.- "incredibly unrealistic, collection size would have to be ___")

contracts/ERC721A.sol
```diff
@@ -85,20 +85,25 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
-            if (ownership.addr != address(0)) {
-                currOwnershipAddr = ownership.addr;
-            }
-            if (currOwnershipAddr == owner) {
-                if (tokenIdsIdx == index) {
-                    return i;
+        uint256 tokenIdsIdx;
+        address currOwnershipAddr;
+
+        // Counter overflow is incredibly unrealistic.
```
chiru-labs, line 91: Please elaborate a bit more on this line (e.g.- "incredibly unrealistic, collection size would have to be ___")

chiru-labs (changes requested):
Sweet stuff. Just some nits

DaniPopes:
Thanks for that! I missed that last one. Also added an unchecked block to `ownershipOf` and added a condition to prevent underflow.

chiru-labs (approved):
Awesome looks good after rebase

cygaar (approved):

cygaar:
Love this

Issue #45, opened 2022-01-29
Question/Change Request: Init `currentIndex` with 1 instead of 0


Pczek:
We've been using ERC721A in our latest NFT drop. It worked great, however, having the token indexes start 0 caused some edgecases in our other contracts for the token with the id 0.

For example we have a simple staking contract were we map the owner address to their staked token. Each address can only stake a single token so its a `mapping(address => uint256) ownerToToken`. However, each address of the map get's initialized with `0` by solidity afaik. So doing stuff like `require(ownerToToken[msg.sender] == 0, "Already staked a token");` does not work properly when supporting the token id 0 in the ERC721A contract.

chiru-labs:
Yup this is on the roadmap. Feel free to open a PR if you have one ready!

Pczek:
👍 Started working on a PR.

Pczek:
Will leave the PR link here as a reference. It's not finished yet. Had only limited time today.
Let me know what you think about the renaming of the variable and it's use.

https://github.com/chiru-labs/ERC721A/pull/66

Pczek:
PR is ready for review. ✌️ 
https://github.com/chiru-labs/ERC721A/pull/66

ottogutierrez:
Quick question about this, if this PR gets approved, the default behaviour would be for the index to start at 1 right?

VRPunk9036:
Wow great work @Pczek , I will do some tests with this to use on one of my projects! Collections starting with 1 are a lot less confusing

Pczek:
> Quick question about this, if this PR gets approved, the default behaviour would be for the index to start at 1 right?

Yes. Having token id `0` can lead to confusion in smart contracts as the initial value for `uint` is `0`.

jpegdigital:
Does function tokenByIndex(uint256 index) need to be adjusted? Token at index 0 is now tokenId 1.

Something like `return index + 1`.

Pczek:
@jpegdigital yes you are right. I've adapted it in my PR.

Fixing this, i've stumbled over the `tokenOfOwnerByIndex` function. Which also seems to need an adaptation. I've added a test for it in my PR however, it's failing and I'm not yet sure what do adapt. Not yet had the time to wrap my head around the `ownership` mapping in the code. 🤔 

https://github.com/chiru-labs/ERC721A/blob/main/contracts/ERC721A.sol#L85-L108

Pczek:
[...]

PR #66, opened 2022-02-02
Start Sequentially Minting at `_startTokenId()` instead of hardcoded 0

Pczek:
As described in #45  NFTs with token ID `0` can cause some headaches for developers down the line. 

I renamed the `currentIndex` to `_nextTokenId` as this describes its use better as well it's an internal variable. But let me know your thoughts about it.
During development I stumbled over this function declaration 
`function getOwnershipAt(uint256 index)` in `ERC721AExplicitOwnershipMock.sol`
For me it was a bit misleading. I expected `_ownership` to be an multidimensional array containing the history of the ownership of an NFT. However, it's a flat mapping between basically the `tokenId` to the `TokenOwnership`. I would suggested renaming the `index` parameter to `tokenId`.
https://github.com/chiru-labs/ERC721A/blob/main/contracts/mocks/ERC721AExplicitOwnershipMock.sol#L19-L21


MrMcGoats:
You should probably add `.vscode` to the git ignore

Pczek:
@MrMcGoats Done. Thanks for pointing that out.

I've added new tests as well. One for `tokenOfOwnerByIndex` is currently failing. 
Any suggestions how to adapt it?

https://github.com/chiru-labs/ERC721A/blob/main/contracts/ERC721A.sol#L85-L108

Pczek:
Fixed the `tokenOfOwnerByIndex` function now as well. I wrote some test for the. The PR is ready for review, happy to get some feedback and merge it if approved. 👍

jpegdigital:
Hi, thought of one more edge case with _exists(). I think `return tokenId < currentIndex;` fails when checking tokenId 0 or the last tokenId? Maybe something like this to fix `return tokenId > 0 && tokenId <= currentIndex;`

I'm surprised there aren't existing tests already that fail for the boundaries. Haven't had a chance to run the test suite though.

Vectorized:
This should have the flexibility to allow users to choose between zero and one indexed tokenIds, with minimal additional gas cost and minimal API change.

contracts/ERC721A.sol
```diff
@@ -60,21 +60,22 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
         _symbol = symbol_;
+        _nextTokenId = 1;
     }
 
     /**
      * @dev See {IERC721Enumerable-totalSupply}.
      */
     function totalSupply() public view override returns (uint256) {
-        return currentIndex;
+        return _nextTokenId - 1;
```
Vectorized, line 70: Can be unchecked.

contracts/ERC721A.sol
```diff
@@ -60,21 +60,22 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
     }
 
     /**
      * @dev See {IERC721Enumerable-tokenByIndex}.
      */
     function tokenByIndex(uint256 index) public view override returns (uint256) {
         require(index < totalSupply(), 'ERC721A: global index out of bounds');
-        return index;
+        return index + 1;
```
Vectorized, line 78: Can be unchecked.

contracts/ERC721A.sol
```diff
@@ -60,21 +60,22 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
     constructor(string memory name_, string memory symbol_) {
         _name = name_;
         _symbol = symbol_;
+        _nextTokenId = 1;
```
Vectorized, line 63: Should not be hardcoded. Users should have the flexibility to choose if they want to start from 0 or 1, minimally.

I've written a version that uses a function override at https://gist.github.com/Vectorized/e6f47b0c1de008bb5983852224e41f69.

contracts/extensions/ERC721AOwnersExplicit.sol
```diff
@@ -13,18 +13,21 @@ abstract contract ERC721AOwnersExplicit is ERC721A {
      */
     function _setOwnersExplicit(uint256 quantity) internal {
         require(quantity != 0, 'quantity must be nonzero');
-        require(currentIndex != 0, 'no tokens minted yet');
+        require(_nextTokenId > 1, 'no tokens minted yet');
```
Vectorized, line 16: Can be changed to `!=` to save tiny gas https://github.com/chiru-labs/ERC721A/pull/59. `_nextTokenId` won't go below 1 anyway.

contracts/extensions/ERC721AOwnersExplicit.sol
```diff
@@ -13,18 +13,21 @@ abstract contract ERC721AOwnersExplicit is ERC721A {
...
+        if (nextOwnerToExplicitlySet == 0) {
```
Vectorized, line 17: Should init nextOwnerToExplicitlySet in constructor.

Vectorized, line 17: Should check and set the _nextOwnerToExplicitlySet local variable instead for gas savings. 

See https://github.com/chiru-labs/ERC721A/pull/59

Pczek:
@Vectorized Thanks for the review. The change requests check out, will incorporate them probably today. Not sure what you mean by `Can be unchecked.`? I guess it relates to the option to make the starting index flexible?
I wonder if you think it should start at 0 or 1 or if even other starting indexes should be valid as well. 

@jpegdigital good call, I've added some `_exists` tests and adapted the function my changes yesterday already. ✌️ 
https://github.com/chiru-labs/ERC721A/pull/66/files#diff-833552e5bd4a4be3ada89b7874eedc6762fb91ed0890dda22e82a1936e6d60d8R280-R282

Vectorized:
> @Vectorized Thanks for the review. The change requests check out, will incorporate them probably today. Not sure what you mean by `Can be unchecked.`? I guess it relates to the option to make the starting index flexible?

Wrap those lines in `unchecked {}` blocks for gas savings, since they cannot overflow and underflow.

See: https://gist.github.com/Vectorized/e6f47b0c1de008bb5983852224e41f69

Pczek:
@Vectorized thanks for the gist, it helped a lot. I've adapted my fork to your suggestions. Works fine now. It needed minor adjustments. 
I've brought back the original test files and plainly copied them for new mock contracts which uses the override to set the starting token id to one. Not sure how I feel about the copy-pasted code, not sure how to do it more elegantly atm.

contracts/ERC721A.sol
```diff
@@ -90,7 +102,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
 
         // Counter overflow is impossible as the loop breaks when uint256 i is equal to another uint256 numMintedSoFar.
         unchecked {
-            for (uint256 i; i < numMintedSoFar; i++) {
+            for (uint256 i = _startTokenId(); i < numMintedSoFar + _startTokenId(); i++) {
```
Vectorized, line 105: It may be more gas efficient to compare it to a local copy of `_nextTokenId`.

```
uint256 end = _nextTokenId;
...
unchecked {
             for (uint256 i = _startTokenId(); i < end; i++) {
```

Vectorized:
@Pczek 

Thanks for your hard work. I was actually preparing to work on this, until I saw your PR. 
Don't worry about copying. Just want to contribute. 

We rise together.
We build together.
We grow together.

TimTinkers:
Likewise, thanks @Pczek.

ottogutierrez:
Thanks @Pczek! I'm excited to see this one on the next release :)

contracts/ERC721A.sol
Pczek: Here are the results in gas for this change. I can only spot a difference in the gas used during deployment. At least it's saving. But seems only to be the case if another `_startTokenId` than `0` is used. 

I've committed this change. 👍 

# Before with `numMintedSoFar`
Screenshot 2022-02-10 at 10 00 44


# After with `end`
Screenshot 2022-02-10 at 10 02 36

Pczek:
@Vectorized @chiru-labs Who can do the final review and approve it to be merged? 
IMHO the PR is ready to be merged.

cygaar:
Once we push the burn function in, I'll review this one :)

Thanks for all the hard work and bearing with us

Vectorized:
@Pczek I've opened a PR to your fork with the added burn functionality. https://github.com/Pczek/ERC721A/pull/1

Pczek:
@Vectorized @cygaar merged https://github.com/Pczek/ERC721A/pull/1 ✅ 
All tests succeed ✅

cygaar:
Awesome thanks, this is the next PR on the docket

Pczek:
@cygaar that would be great! I've just resolved the latest conflict and adapted the code of mine to work with the recent changes of the `main` branch.

Regarding a gas report. Here's a screenshot of the current `main` and my fork side by side. The only difference I'm able to spot is the cost of deployment. However, these are expected as more code is deployed. 🤷 


Screenshot 2022-02-18 at 14 03 48

Vectorized:
Interesting how the mint actually used lower gas on average when the starting index is 1 instead of 0. 

When the index starts at 0, the very first mint will change `_currentIndex` from 0 to 1, which costs more gas (20000 vs 5000).

ahbanavi:
@Pczek I've opened a PR https://github.com/Pczek/ERC721A/pull/3 to your branch for adding `_totalMinted` function (#124).

cygaar:
Thanks for your PR @ahbanavi let's wait for that to get merged and then I'll do an in-depth review of this PR

Pczek:
@Vectorized @ahbanavi just merged your PR https://github.com/Pczek/ERC721A/pull/3

Test seem to run fine, I haven't adapted anything else since the PR of @ahbanavi .

Anything left to do? I saw you opened up a new PR @Vectorized ?

Vectorized:
@Pczek Welcome back! 

I'll close that PR first. Let's wait for @cygaar.

cygaar:
Going to review this today

contracts/mocks/ERC721AStartOneMock.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT
+// Creators: Chiru Labs
+
+pragma solidity ^0.8.4;
+
+import '../ERC721A.sol';
+
+contract ERC721AStartOneMock is ERC721A {
```
cygaar, line 8: You should just inherit from `ERC721AMock` instead of redeclaring everything again

contracts/mocks/ERC721AStartOneExplicitOwnershipMock.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+// Creators: Chiru Labs
+
+pragma solidity ^0.8.4;
+
+import '../extensions/ERC721AOwnersExplicit.sol';
+
+contract ERC721AStartOneOwnersExplicitMock is ERC721AOwnersExplicit {
```
cygaar, line 8: Inherit from `ERC721AOwnersExplicitMock` instead of duplicating everything

test/ERC721AStartOne.test.js
```diff
@@ -0,0 +1,341 @@
+const { expect } = require('chai');
+const { constants } = require('@openzeppelin/test-helpers');
+const { ZERO_ADDRESS } = constants;
+
+const RECEIVER_MAGIC_VALUE = '0x150b7a02';
+const GAS_MAGIC_VALUE = 20000;
+
+describe('ERC721AStartOne', function () {
+  beforeEach(async function () {
+    this.ERC721A = await ethers.getContractFactory('ERC721AStartOneMock');
```
cygaar, line 10: I think a better approach to our tests would be parameterize this class variable to change between the 0-index and 1-index contracts so we don't have to duplicate most of the code.

You would then have specific tests for `ownerOf` and tokenId. You would probably need a separate context for 1-index specific test cases.

I want to minimize the work we need to do anytime we add/change some functionality in our contract.

test/ERC721A.test.js
```diff
@@ -18,6 +18,11 @@ describe('ERC721A', function () {
       const supply = await this.erc721a.totalSupply();
       expect(supply).to.equal(0);
     });
+
+    it('has 0 totalMinted', async function () {
```
cygaar, line 22: Can you add totalMinted test cases for ERC721ABurnable?

cygaar (changes requested):
The actual code looks fine, @Vectorized I think we can make the readme/comment changes in another PR. I'm more concerned about test quality and reducing copy-pasted code

test/ERC721A.test.js
ahbanavi: I add one here:  
https://github.com/chiru-labs/ERC721A/blob/f0a5bb42f85300e9857932c5ddb79164dd07b124/test/extensions/ERC721ABurnable.test.js#L56-L63  
Are there any other situations?

Vectorized: This is hard. 

I'm thinking of simply adding a `directBurn` function to `ERC721AMock` if we want to test `_totalMinted` with `_burn` and `_startTokenId`. 

The route of having a `ERC721ABurnableStartOne`, or making `ERC721AMock` subclass `ERC721ABurnable` feels uncomfortable.

test/ERC721AStartOne.test.js
Pczek: Agree, to remove most of the duplicate test code. This is a good hint, I will look into that. When I copied them I didn't know better. Will try to adapt it today.

Pczek:
@Vectorized @cygaar please check my recent changes to make a parameterized test. How do you like it this way?
If this is fine for you we can adapt the rest of the duplicated tests for the ExplicitOwnership extension as well.

[`2ba810d` (#66)](https://github.com/chiru-labs/ERC721A/pull/66/commits/2ba810d4c241b36b6c82bca122c65a6ba699ae6c#diff-e166966f732548cffaf8c83a2732c01fdf82cad2d861478e0b1dd43ba00d9fe2)

contracts/mocks/ERC721AStartTokenIdMock.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+// Creators: Chiru Labs
+
+pragma solidity ^0.8.4;
+
+import './ERC721AMock.sol';
+
+contract StartTokenId {
+    uint256 startTokenId;
```
Vectorized, line 9: Make `startTokenId` immutable for zero-cost abstraction. Otherwise, each call to _startTokenId will incur gas. 

You can also make this public and remove the `getStartTokenId` function.

Consider moving this class to a separate file, since you are going to subclass it from `OwnersExplicit` too.

cygaar:
LGTM. @Pczek parameterize the ownership explicit test and this is good to go. Appreciate all the work you and everyone has put in to get this out

test/ERC721A.test.js
Pczek: Would be cool to test these features individually, however, these are quite connected. 
So I created a new mock combining both in one of my recent commits:
[`e16a79a` (#66)](https://github.com/chiru-labs/ERC721A/pull/66/commits/e16a79a5bc8382aef740cc843e3c4d04a3ff03d9)

Pczek:
Alright, I've parameterized now the following tests:
`ERC721A.test`
`ERC721AOwnersExplicit.test`
`ERC721ABurnable.test`

All of them now also run their test suite for a contract which is modified with the `StartTokenIdHelper.sol` contract to set the override the `_startTokenId()` function dynamically.

Afaik no further tasks are open on this PR until another review or it's ready to be merged.

CC @Vectorized @cygaar

test/ERC721A.test.js
```diff
@@ -5,319 +5,354 @@ const { ZERO_ADDRESS } = constants;
...
+            await expect(this.erc721a.mint(this.owner.address, 0, data, false)).to.be.revertedWith('MintZeroQuantity');
+          });
+        });
       });
     });
-  });
-});
+  };
+describe('ERC721A', createTestSuite({ contract: 'ERC721AMock', constructorArgs: ['Azuki', 'AZUKI'] }));
+describe(
```
cygaar, line 355: Styling nit: can you add a new line between lines 354 and 355?

cygaar (approved):

test/ERC721A.test.js
Pczek: Sure done ✅  Good catch 👍

PR #73, opened 2022-02-04
feature: use custom errors

qd-qd:
## Context

Starting from Solidity [v0.8.4](https://blog.soliditylang.org/2021/04/21/custom-errors/), there is a convenient and gas-efficient way to explain to users why an operation failed through the use of custom errors. This way is less expensive, especially when it comes to deploying cost than passing a string to require or assert.

## Gas comparison 

**Before**
before

**After**
after

## Result

As you can observe, the deployment cost is highly reduced for the mocked contracts (almost 20%).

## Tradeoff

This improvement requires a bump to 0.8.**4**.  I am personally convinced this is worth it and the accessibility issue caused by the bump is acceptable due to the fact the bump is minor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ npm install --save-dev erc721a
 Once installed, you can use the contracts in the library by importing them:
 
 ```solidity
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.4;
 
 import "erc721a/contracts/ERC721A.sol";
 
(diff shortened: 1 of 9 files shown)


cygaar:
my biggest concern is that we want to match open zeppelin's accessibility, and they still have their library at ^0.8.0

cygaar:
It's not the biggest deal in the world, I'll talk to the rest of the team and see what they think.

cygaar:
Alright, let's do this. Thanks for making this change

cygaar (approved):

ahbanavi:
Thanks for this PR, Wouldn't it better that we use NatSpec to further explain the reasons behind the error for convenient error-reporting?  
If it does, I can create another PR with additional changes.

PR #74, opened 2022-02-04
docs: fix the command to run hardhat node

qd-qd:
`npm node` won't work, we need to use `npm run node` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,7 @@ Don't forget to give the project a star! Thanks again!
 ### Running tests locally
 
 1. `npm install`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,7 +83,7 @@ Don't forget to give the project a star! Thanks again!
 ### Running tests locally
 
 1. `npm install`
-2. `npm node`
```
cygaar, line 86: we should just remove 3, and make 2 `nom run test`

cygaar (changes requested):

PR #86, opened 2022-02-09
add _unsafeMint function

ahbanavi:
For easier use so that we do not have to fill all 4 parameters.
I think with that way, mocking and testing `_mint` function is redundant, let me know if it did, so I remove related tests.  

Also, I think `_mint` function should be `private` instead of `internal` cause with `_safeMint` and `_unsafeMint` functions there's no additional need for developers to use this method directly. I wait to hear other's opinion about this.

I will add testGas tests after I figure out whether we should make `_mint` function internal or make it private.  Also, there are some bugs in [test/GasUsage.test.js](https://github.com/chiru-labs/ERC721A/blob/7b3038e294253e75b4878a530c8c9a8ec293f0a8/test/GasUsage.test.js) context names, I'm not sure to fix them in this PR or make another.

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -305,6 +305,11 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
      * - If `to` refers to a smart contract, it must implement {IERC721Receiver-onERC721Received}, which is called for each safe transfer.
      * - `quantity` must be greater than 0.
      *
+     * Requirements:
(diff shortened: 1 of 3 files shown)


ahbanavi:
Makes sense, I open another PR later for fixes on [test/GasUsage.test.js](https://github.com/chiru-labs/ERC721A/blob/7b3038e294253e75b4878a530c8c9a8ec293f0a8/test/GasUsage.test.js).

PR #89, opened 2022-02-10
Optimize approval check

Vectorized:
**Rationale**

`setApprovalForAll` is typically used more often than `approve`. (e.g. on marketplaces).

In practice, `isApprovedForAll` is more likely to return true compared to  `getApproved`.

Moving the `isApprovedForAll` check before `getApproved` can save some gas via boolean short circuiting.


ahbanavi (approved):
Looks good and makes sense

cygaar (approved):

PR #95, opened 2022-02-12
Fix context name and method called not coincident.

themez:


ahbanavi (changes requested):
LGTM, and I think you miss one line:

themez:
> LGTM, and I think you miss one line at [L36](https://github.com/chiru-labs/ERC721A/pull/95/files#diff-c18f7307065b27faeb077a4a82d34211a57f8956671df987409b0bedae862529R36)

Thanks, I've updated it accordingly.

cygaar (approved):
Good catch

PR #98, opened 2022-02-13
Add aux to AddressData

Vectorized:
After adding the `burn` feature, the `AddressData` struct still has 64 bits of unused space.

**_This is free real estate._** 

This PR adds an `aux` variable for the `AddressData` to let users make full use of the extra space,   
along with the relevant `_getAux` and `_setAux` functions.

This `aux` can be used for storing miscellaneous things like number of whitelist mint slots used.

Deriving contracts can avoid the overhead of an additional expensive SLOAD and SSTORE with this feature.

If there are no likely plans to use the extra 64 bits of space for future features, we can consider this.

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -19,6 +19,7 @@ error ApprovalToCurrentOwner();
 error BalanceQueryForZeroAddress();
 error MintedQueryForZeroAddress();
 error BurnedQueryForZeroAddress();
+error AuxQueryForZeroAddress();
 error MintToZeroAddress();
 error MintZeroQuantity();
 error OwnerIndexOutOfBounds();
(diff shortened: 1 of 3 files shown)


ahbanavi (approved):
Looks good and yes it's the perfect place for storing whitelist mints.

cygaar:
Good find

cygaar (changes requested):
Can you add a couple unit tests to verify the behavior? Should be simple since you're just checking that the setter/getter works.

cygaar:
@Vectorized for some reason you have changes from the past 2 merged PRs in your PR. It might be worth rebasing your code on top of main instead of trying to do a merge commit. Not too sure what happened. Other than that, the PR looks good

PR #100, opened 2022-02-14
Update projects.md to add Capsule Machine NFT

CapsuleMachine:
Capsule Machine NFT uses ERC721A

https://etherscan.io/address/0xc19ced6633f0da7cef642b7a3f6b3ff0bb2465c0#code


CapsuleMachine:
Should the list be modified to be in chronological order per transaction time of the contract being created?

cygaar:
We might reorganize it in the future, but for now it's FCFS

PR #106, opened 2022-02-14
Add Tokemon Cards to projects.md

tokemoncards:
Add Tokemon Cards


cygaar:
I'm only going to merge PRs with deployed contracts, so feel free to open this when you have that

tokemoncards:
Thank you so much for the update @cygaar - we will re-open once we have deployed

Just a heads up it looks like there are 7 or so projects that made it through without a deployed contract link in case you wanted to do some housecleaning

PR #107, opened 2022-02-14
Remove tokenOfOwnerByIndex

### LINKED ISSUE #69: Move IERC721Enumerable impl to an extension
I don't think everyone needs this on-chain, and considering the risks of DOS related to an O(totalSupply) method (tokenOfOwnerByIndex), I do think it'd make a lot of sense to have this implemented in an extension.

fulldecent:
Fixes #69 (ish)

tokenOfOwnerByIndex is incompatible with ERC721a. This PR removes it.

Therefore we lose ERC721Enumerable extension. That's fine.

But we also keep tokenbyIndex and totalSupply. It is fine to implement these functions even though we are not implementing ERC721Enumerable. Nobody says we can't do this.

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -40,7 +40,7 @@ error URIQueryForNonexistentToken();
  *
  * Assumes that the maximum token id cannot exceed 2**128 - 1 (max value of uint128).
  */
-contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable {
+contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
     using Address for address;
     using Strings for uint256;
 
@@ -100,7 +100,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
     /**
      * @dev See {IERC721Enumerable-totalSupply}.
      */
-    function totalSupply() public view override returns (uint256) {
+    function totalSupply() public view returns (uint256) {
         // Counter underflow is impossible as _burnCounter cannot be incremented
         // more than _currentIndex times
         unchecked {
@@ -113,7 +113,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
      * This read function is O(totalSupply). If calling from a separate contract, be sure to test gas first.
      * It may also degrade with extremely large collection sizes (e.g >> 10000), test for your use case.
      */
-    function tokenByIndex(uint256 index) public view override returns (uint256) {
+    function tokenByIndex(uint256 index) public view returns (uint256) {
         uint256 numMintedSoFar = _currentIndex;
         uint256 tokenIdsIdx;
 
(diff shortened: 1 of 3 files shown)


Vectorized:
I think `tokenByIndex` should be removed too for this PR, it's O(totalSupply) with burn function.

The line containing `IERC721Enumerable` in the `supportsInterface` function can be removed too.
https://github.com/chiru-labs/ERC721A/blob/34788d87ef3dd16f54ce82b5b038966bff1d64ec/contracts/ERC721A.sol#L143

fulldecent:
Yes, removed `tokenByIndex`.

Kept the inheritance in order to qualify this reference in `totalSupply`: `See {IERC721Enumerable-totalSupply}`.

syffs:
@fulldecent I did this #81 a week ago.

fulldecent:
Right but #81 adds a new extension. I don't think that new extension should exist.

syffs:
> Right but #81 adds a new extension.

It moves IERC721Enumerable to an extension, so that ERC721A is free of it.

>  I don't think that new extension should exist.

Why ? There're very selected cases in which ERCEnumerable functions may be useful on-chain.

Always been the same deal with OZ: they moved IERCEnumerable because they couldn't assume that it would never be make sense... 
Maybe it should include very clear warnings though: so many devs who don't know better still use OZ ERC721Enumerable even though OZ themselves recommand not to use it.

fulldecent:
OpenZeppelin contracts includes ERC721Enumerable because it is a feasible extension. It is implemented as O(1).

ERC721Enumerable is not O(1), therefore ERC721Enumerable is not feasible here.

I think the threshold for this repository should be/is: "things that people probably want with ERC721A". Not "dangerous things that we have to warn people not to use and that nobody specifically wants and which are basically not compatible with ERC721A". Therefore the extension should not be included here.

But again, this repo is missing a mission scope, so I cannot argue further on this topic without any source of authority.

cygaar:
I'm in agreement with @fulldecent here - these functions were originally implemented to market gas improvements, but ERC721A has become primarily a means of minting tokens in batch efficiently. At the current point in time we want to make this contract as safe as possible and prevent someone from doing something dangerous.

We will update the README with a better mission statement shortly

cygaar (approved):
Approving this, but will wait a day or so before merging so our community members can add their thoughts

syffs:
> OpenZeppelin contracts includes ERC721Enumerable because it is a feasible extension. It is implemented as O(1).

this "feasibility" you mention is highly debatable. Only that the issue isn't security related, but gas related: devs who don't know better keep implementing this extension that takes ~2M gas for minting 20 tokens for no functional reason in most cases. Now this is not a subject for this repo.

Anyway, this is a bit disappointing that someone put the work into changing this as requested [here](https://github.com/chiru-labs/ERC721A/pull/29#issuecomment-1023513507), and that you would choose to submit a new PR instead of commenting an existing one...

FYI this line should be removed:
https://github.com/chiru-labs/ERC721A/blob/584d0735b8eab3e5e61ceaaccc1931450b93de2d/contracts/ERC721A.sol#L9

fulldecent:
Developers /do/ know better and they have /chosen/ to implement ERC721Enumerable. Very many projects do this. There is nothing wrong this this extension. The tradeoffs have been discussed thoroughly in the [standardization process](https://eips.ethereum.org/EIPS/eip-721) and this is implemented in [the first ERC-721 token](https://tenthousandsu.com).

Arguing this in either direction probably won't go anywhere. Especially against me.

---

For OpenZeppelin Contracts, a good implementation of a well-known standard, following the specifications, and achieving O(1) complexity is in scope for that project. Their implementation meets those requirements, that's why it's published in that project.

---

If you want to feel better, watch me embarrass myself by developing a huge PR with the team and then the /real/ decision maker decides they don't have time to review PRs any more.

https://github.com/google/open-location-code/pull/463/

I have piles of examples like this.

syffs:
> Arguing this in either direction probably won't go anywhere. Especially against me.

I'll agree with you on this: it is completely pointless !

> There is nothing wrong this this extension

You don't know better is all: have a chat with OZ devs before you mention it

> If you want to feel better, watch me embarrass myself by developing a huge PR with the team and then the /real/ decision maker decides they don't have time to review PRs any more.

Nothing closely related to this situation ! This is not about you: open source is about collaboration, disrespecting other's work to flatter your ego certainly isn't the spirit.

Wasted enough of my time already, I'm done here.

fulldecent:
If there are unresolved criticisms here of the OpenZeppelin Contracts `ERC721Enumeration` implementation, please open the issue there and feel free to mention me. I'm happy to learn more on this topic.

Also I just want you to know I appreciate this discussion here and that PR that's being discussed. Arguing about a PR's acceptability doesn't change that.

Vectorized:
Just leaving this here for those interested.

https://www.twitch.tv/videos/1299266697 <- (around 48 minute mark) 
^ The steam will only be available for 2 weeks after recording.

I've asked William about his thoughts on ERC721A. Specifically on the enumerable parts, and whether we should include a convenience `tokensOfOwner` function for small collections (he recommends against it). 

Designing standards / libraries to be efficient, future proof, and generalizable requires a lot of thought.

ahbanavi (commented):
Shouldn't this line be deleted in `supportsInterface`?
https://github.com/chiru-labs/ERC721A/blob/f3205a76816fc6fed1a0090d9be5d34b30bb1485/contracts/ERC721A.sol#L175

Issue #124, opened 2022-02-19
feat: `_totalMinted()` function.


Vectorized:
With the burn feature, collections will need to check against `_currentIndex` if they want to limit the total number of mints.
Checking against `totalSupply()` can result in new mints being possible if someone burns their token(s).

With upcoming PR #66, users will have to check against `_currentIndex - _startTokenId()` instead.

This can cause some confusion.

I suggest adding an internal `_totalMinted()` function,  
and update the README to encourage checking against `_totalMinted()` instead.

Another option is to put this into #66. 

Let me know your comments. 

If you want to make a PR for this issue, please feel free to do so. Drop a comment below too. 

Otherwise, I will probably wait a few / several days after #66 is merged before starting a PR for this (if this feature is not in a PR or merged by then). 

@cygaar @ahbanavi

ahbanavi:
Yes, `_totalMinted()` method would be convenient for this.  
IMO, we better wait for #66 to be merged first. 
Putting it in #66 also do the job I think, and it would be faster, but it's out of context of that PR.

cygaar:
I think I'd prefer it in #66 just so no one accidentally copies this code and runs into an issue checking against `_currentIndex`. Thoughts on that approach? I'd like to make that change "atomic"

ahbanavi:
I agree to add in #66. True, being “atomic” is important on this matter.

PR #129, opened 2022-02-23
Add customizable `_startTokenId()` and `_totalMinted()` (Continued from #66)

Vectorized:
Continued from #66. Kept the history of commits to preserve contributions.

Summary:
- Enables the `_startTokenId()` to be customized via function override.
  - Prevents the need for a 3rd constructor argument, which can confuse users, as earlier versions of ERC721A used it for the `maxBatchSize`.
  - Compilers >= 0.8.2 will automatically inline the calls for zero-cost abstraction. 
  Not a problem as we only support >= 0.8.4.
- Includes an internal `_totalMinted()` function. See #124 
  - Internal because it is not part of the standards, and we want to give users the choice to let it remain internal / wrap it in a public function.
  - Edited `README` to tell people to use `_totalMinted()`. 

Some quality of life suggestions (not yet added), please drop your opinions below 
(include it in this PR / move to future PR / not needed / etc.):
- Prettify the code to a max line length of 99. I know most of the long lines are from OZ. 
But that doesn't mean we can't prettify things a bit.
- Add a `_mint(to, quantity)` function. 
This for the `README` example.
- Add a cautionary comment to `_safeMint` that it is not re-entrancy safe, and encourage users to use `_mint` instead?
This is for the `README` example.

Many projects are using ERC721A. Many of which are not listed. 
Our contributions do have a huge impact, probably more than we can imagine.

@cygaar @ahbanavi @Pczek

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,6 @@ contract Azuki is ERC721A {
 
 ## Roadmap
 
-- [] Add flexibility for the first token id to not start at 0
 - [] Support ERC721 Upgradeable
 - [] Add more documentation on benefits of using ERC721A
 - [] Increase test coverage
--- a/README.md
+++ b/README.md
@@ -40,11 +40,17 @@ pragma solidity ^0.8.4;
 import "erc721a/contracts/ERC721A.sol";
 
 contract Azuki is ERC721A {
+
+  uint256 public constant MAX_SUPPLY = 10000;
+
   constructor() ERC721A("Azuki", "AZUKI") {}
 
   function mint(uint256 quantity) external payable {
-    // _safeMint's second argument now takes in a quantity, not a tokenId.
-    _safeMint(msg.sender, quantity);
+    // Use `_totalMinted()` to get the number of tokens minted.
+    require(_totalMinted() + quantity <= MAX_SUPPLY, "Sold out!");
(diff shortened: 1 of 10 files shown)


ahbanavi:
Thanks for opening this PR.  

My opinion about suggestions:  
> Prettify the code to a max line length of 99. I know most of the long lines are from OZ.
> But that doesn't mean we can't prettify things a bit.

Agree to use [prettier](https://prettier.io/) for both `solidity` and `js`.  
I have to manually disable my `prettier` before my contributions because it changed a lot of lines in the whole project, and it would be better if everyone were using this.
Also, It might help if write a ` Contributing guidelines` and tell people to use `prettier` for their contribution.  
But it's better we do this on a future PR.  

> Add a _mint(to, quantity) function.
> This for the README example.

I opened a PR before about adding `_unsafeMint` #86, check it out.  

>Add a cautionary comment to _safeMint that it is not re-entrancy safe, and encourage users to use _mint instead?
> This is for the README example.

Using `_mint` function without checking on `_checkOnERC721Received` may also cause problems, I mean developers should know the differences between safe and normal mint.  
But it doesn't hurt to give some information on this matter in `README` and explain the different approaches (like using normal mint if their already revert requests from contracts).  
This could be on another PR too, maybe the same PR for prettifying.

Issue #92, opened 2022-02-11
Possible reentrancy issues in _mint() function?


jpegdigital:
Hi, I'm looking to implement ERC721A in a project so have been combing through the code. Still new to reentrancy patterns so wanted to ask the question first before submitting a PR.

When calling _safeMint(), control is handed off to a receiving contract before totalSupply is updated. Could a malicious contract create an exploit or wreck havoc by re-entering the mint function from here? In the re-entrant call, startTokenId would be identical to the original call, but balances changed and events have been emitted so it's not entirely obvious to me what kind of damage could occur.

Code in question:
```
uint256 startTokenId = _currentIndex;
...
unchecked {
    _addressData[to].balance += uint128(quantity);
    _addressData[to].numberMinted += uint128(quantity);

    _ownerships[startTokenId].addr = to;
    _ownerships[startTokenId].startTimestamp = uint64(block.timestamp);

    uint256 updatedIndex = startTokenId;

    for (uint256 i; i < quantity; i++) {
        emit Transfer(address(0), to, updatedIndex);
        if (safe && !_checkOnERC721Received(address(0), to, updatedIndex, _data)) {
            revert TransferToNonERC721ReceiverImplementer();
        }

        updatedIndex++;
    }

    _currentIndex = updatedIndex;
}

```

jpegdigital:
In terms of solutions/alternatives, I've been thinking of three options for my project:

1. Remove _safeMint() all together as handing off control to another contract is risky in this 2022 minting environment.
2. Call _checkOnERC721Received() only once for the final token (ensures state is fully updated). Or maybe creating a new _checkOnERC721ReceivedBatch standard.
3. Ignore and advise implementing contracts to use reentrancy blockers.

Vectorized:
Good points. Looks like balance and the number minted can become inconsistent.

Perhaps a simple check like this could help:
```
if (safe && _currentIndex != startIndex) revert();
_currentIndex = updatedIndex;
```

The compiler could optimize it away if safe is always false, but otherwise the gas overhead should be minimal.

ahbanavi:
I think the implementing contracts should handle reentrancy, as many contracts (like Azuki) block minting from other contracts in the first place and therefore there's no reentrancy problem (correct me if I'm wrong).  
So a warning in `_safeMint` specs or in the docs I think is enough, and we should keep `ERC721A` implementation close to the original.

cygaar:
[...]

PR #131, opened 2022-02-23
Mint re-entrancy guard with optimizations.

Vectorized:
Ok, so as we know, the `_safeMint` function is **NOT** safe from re-entrancy. #92 

Ideally, most contracts should use `_mint`.

Although we can warn about it in the comments / docs, many devs won't read it and will happily use `_safeMint` as it is. 
For security reasons, I won't cite any examples here. 

Furthermore, the README example shows `_safeMint` being used in an unprotected manner. 

From a usability point of view, I suggest that we make the `_mint` function re-entrancy safe regardless, to be foolproof. 

This will also reduce the cognitive dissonance of `_safeMint` being unsafe, giving a cleaner API design.

The overhead is only 100 gas for an extra warm SLOAD. 
Assuming a function using `_safeMint` costs ~60K gas, this is only 0.2% overhead that can protect against catastrophic consequences. For users who use `_mint`, the gas overhead is as good as none.

To ease the concerns for the increase in gas for `_safeMint`, the function is optimized even further, such that we can save more gas to cover for the overhead.

In practice, implementations that need to use `_safeMint` for some reason can save 2k gas (cold SLOAD) with this functionality, as they don't need to use OpenZepplin's Reentrancy guard. 

**Before:**

```
·------------------------------------------|---------------------------|-------------|
|           Solc version: 0.8.11           ·  Optimizer enabled: true  ·  Runs: 800  ·
···········································|···························|·············|
|  Methods                                                                            
···························|···············|·············|·············|·············|
|  Contract                ·  Method       ·  Min        ·  Max        ·  Avg        ·
···························|···············|·············|·············|·············|
|  ERC721AGasReporterMock  ·  mintOne      ·      56502  ·      90702  ·      57186  ·
···························|···············|·············|·············|·············|
|  ERC721AGasReporterMock  ·  mintTen      ·      74696  ·     108896  ·      75380  ·
···························|···············|·············|·············|·············|
|  ERC721AGasReporterMock  ·  safeMintOne  ·      59256  ·      93456  ·      59940  ·
···························|···············|·············|·············|·············|
|  ERC721AGasReporterMock  ·  safeMintTen  ·      79246  ·     113446  ·      79930  ·
···························|···············|·············|·············|·············|
|  Deployments                             ·                                         ·
···········································|·············|·············|·············|
|  ERC721AGasReporterMock                  ·          -  ·          -  ·    1146497  ·
···········································|·············|·············|·············|
```

**After re-entrancy guard + gas optimizations:**
```
·------------------------------------------|---------------------------|-------------|
|           Solc version: 0.8.11           ·  Optimizer enabled: true  ·  Runs: 800  ·
···········································|···························|·············|
|  Methods                                                                            
···························|···············|·············|·············|·············|
|  Contract                ·  Method       ·  Min        ·  Max        ·  Avg        ·
···························|···············|·············|·············|·············|
|  ERC721AGasReporterMock  ·  mintOne      ·      56462  ·      90662  ·      57146  ·
···························|···············|·············|·············|·············|
|  ERC721AGasReporterMock  ·  mintTen      ·      74116  ·     108316  ·      74800  ·
···························|···············|·············|·············|·············|
|  ERC721AGasReporterMock  ·  safeMintOne  ·      59143  ·      93343  ·      59827  ·
···························|···············|·············|·············|·············|
|  ERC721AGasReporterMock  ·  safeMintTen  ·      76775  ·     110975  ·      77459  ·
···························|···············|·············|·············|·············|
|  Deployments                             ·                                         ·
···········································|·············|·············|·············|
|  ERC721AGasReporterMock                  ·          -  ·          -  ·    1168285  ·
·------------------------------------------|-------------|-------------|-------------|
```

@cygaar @ahbanavi

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -399,7 +399,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
             uint256 end = updatedIndex + quantity;
 
             if (safe && to.isContract()) {
-                while (updatedIndex < end) {
+                while (updatedIndex != end) {
                     emit Transfer(address(0), to, updatedIndex);
                     if (!_checkOnERC721Received(address(0), to, updatedIndex, _data)) {
                         revert TransferToNonERC721ReceiverImplementer();
@@ -409,7 +409,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
                 // Reentrancy protection
                 if (_currentIndex != startTokenId) revert();
             } else {
-                while (updatedIndex < end) {
+                while (updatedIndex != end) {
                     emit Transfer(address(0), to, updatedIndex);
                     updatedIndex++;
                 }    
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -396,16 +396,24 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
             _ownerships[startTokenId].startTimestamp = uint64(block.timestamp);
 
             uint256 updatedIndex = startTokenId;
+            uint256 end = updatedIndex + quantity;
 
-            for (uint256 i; i < quantity; i++) {
-                emit Transfer(address(0), to, updatedIndex);
-                if (safe && !_checkOnERC721Received(address(0), to, updatedIndex, _data)) {
-                    revert TransferToNonERC721ReceiverImplementer();
+            if (safe && to.isContract()) {
+                while (updatedIndex < end) {
+                    emit Transfer(address(0), to, updatedIndex);
+                    if (!_checkOnERC721Received(address(0), to, updatedIndex, _data)) {
+                        revert TransferToNonERC721ReceiverImplementer();
+                    }
+                    updatedIndex++;
                 }
-                updatedIndex++;
+                // Reentrancy protection
+                if (_currentIndex != startTokenId) revert();
+            } else {
+                while (updatedIndex < end) {
+                    emit Transfer(address(0), to, updatedIndex);
+                    updatedIndex++;
+                }    
             }
-            // Reentrancy protection.
-            if (safe && _currentIndex != startTokenId) revert();
             _currentIndex = updatedIndex;
         }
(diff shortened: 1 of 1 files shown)


contracts/ERC721A.sol
```diff
@@ -375,15 +375,24 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
...
 
             uint256 updatedIndex = startTokenId;
+            uint256 end = updatedIndex + quantity;
 
-            for (uint256 i; i < quantity; i++) {
-                emit Transfer(address(0), to, updatedIndex);
-                if (safe && !_checkOnERC721Received(address(0), to, updatedIndex, _data)) {
-                    revert TransferToNonERC721ReceiverImplementer();
+            if (safe && to.isContract()) {
+                while (updatedIndex != end) {
```
Vectorized, line 381: The `!=` check is more gas optimized and also doubles as an overflow guard.

Even if the user sets the `_startTokenId()` to a value near `2**256`, it will still work.

If there is overflow, the loop will loop so many times that it gives an out of gas error.

cygaar (approved):
I like it, let's land this after the start index PR like you said

liarco:
Hi, thank you all for the great job with this project.

I just tried to get in touch with @Vectorized from other places but I couldn't, so I hope not to bother you asking this question here.

About what you said here:
> Ideally, most contracts should use _mint.
>
> Although we can warn about it in the comments / docs, many devs won't read it and will happily use _safeMint as it is.

Do you mean that using `_safeMint()` instead of `_mint()` is a bad practice in general, or do you refer to just using it without considering potential re-entrancy attacks?

Would you mind explaining why you would suggest using `_mint()` in most situations?

Thank you for your time.

Vectorized:
@liarco Messaged you on twitter. I forgot to turn on direct messages. You can contact me there from now.

https://ethereum.stackexchange.com/questions/115280/mint-vs-safemint-which-is-best-for-erc721 <- by the lead author of ERC721.

PR #141, opened 2022-02-27
Add Candy Chain to projects.md

kyokosdream:
Adds Candy Chain to the projects list. Need to confirm with authors if Mumbaiscan url is okay with them. Can omit until we publish an Etherscan mainnet version if they'd like.


cygaar (approved):

cygaar:
Actually can you add this when the etherscan is available?

cygaar (changes requested):
.

kyokosdream:
Yes, we will verify on Etherscan today. A user caught an error in one of our require checks so we are just running some tests to make sure it's good to go before we spend currency on deployment and verification.

kyokosdream:
@cygaar Added Etherscan address

Issue #148, opened 2022-03-03
Make setApprovalForAll virtual


KirienzoEth:
Hello!

Would it be possible to make `setApprovalForAll` virtual? That's the case in OpenZeppelin's ERC721 implementation. Or are there any reasons why `isApprovedForAll` is virtual, but not `setApprovalForAll`?

It would be useful because of the various gas optimization strategies involving approving marketplace proxys by default (among others), which require you to override `isApprovedForAll` to add some custom logic.

IMHO, some custom logic is also necessary in `setApprovalForAll` so users can revoke those default approvals the same way they'd revoke any address's approvals, thus why I would like to be able to override that method.

Vectorized:
Sounds good. You can make a PR. 

@cygaar what do you think?

KirienzoEth:
Done here: #149

cygaar:
PR merged

KirienzoEth:
Thank you 🙏

PR #149, opened 2022-03-04
Marks setApprovalForAll as virtual

KirienzoEth:
issue #148


cygaar:
Thanks for doing this

PR #160, opened 2022-03-09
Removed zero address check on _numberMinted

BenSparksCode:
There is already a zero address check on `mint`, so you can safely check the zero address via this view function - it should always return zero.

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -152,7 +152,6 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
      * Returns the number of tokens minted by `owner`.
      */
     function _numberMinted(address owner) internal view returns (uint256) {
-        if (owner == address(0)) revert MintedQueryForZeroAddress();
         return uint256(_addressData[owner].numberMinted);
     }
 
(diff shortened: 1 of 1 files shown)


Vectorized:
Nice find! Love it. 🧠

ahbanavi:
Should we consider removing `BurnedQueryForZeroAddress` or `AuxQueryForZeroAddress` too?

Vectorized:
> Should we consider removing `BurnedQueryForZeroAddress` or `AuxQueryForZeroAddress` too?

@ahbanavi Yes. Wow, you can read my mind.

Vectorized:
Going to merge this after #165. 

@ahbanavi We can make a separate PR for the other removals instead.

ahbanavi:
@Vectorized Other removals can wait for another PR.  
But IMO
https://github.com/chiru-labs/ERC721A/blob/2fc12184a02cd211fbc6472564ad1d7c0bc07c9f/contracts/ERC721A.sol#L20
should be removed in this one.

BenSparksCode:
@ahbanavi @Vectorized  done :)

PR #161, opened 2022-03-10
chore: unify internal function style

antoncoding:
All internal functions are prefixed with `_` except `ownershipOf`. Should be slightly better to have unified style.


ahbanavi (changes requested):
Good catch, but IMHO you should remove `package-lock.json` from the changes.

ahbanavi (approved):

PR #174, opened 2022-03-14
No need to set the slot of tokenId+1 explicitly when burn

xinyang12:


ahbanavi:
Why? Can you share more information?

Vectorized:
Very interesting observation. 

The idea is that _ownershipOf will still find the correct TokenOwnership, although it has to keep iterating all the way to the start of the batch.

One possible issue is that if the batch size is too big, burning without initializing slots can lead to very expensive total cost. 

I'm thinking perhaps we can skip the initialization for all token IDs except for `tokenId & 7 == 0`.

xinyang12:
Deleting these lines has nothing to do with the **burned** status of the next Id.
But it will make the **startTimestamp** uncorrect. you are right.

PR #180, opened 2022-03-16
Add ERC721AUpgradeable

iainnash:
This follows previous attempts at adding an openzeppelin style `ERC721AUpgradeable` contract using the latest version of ERC721A.

Would be interesting to have a script that copies the body of the contract between the ERC721A and ERC721AUpgradable implementations but for now this is a reasonable approach.

All tests pass.


iainnash:
Another approach could be to create a separate repo that uses a build script to pull down the ERC721A Repo and does these changes automatically simply through some sort of find/replace and add.

Vectorized:
Thank you for your submission.

We have a transpiler based implementation in the works.

https://github.com/Vectorized/ERC721A-Transpiler
https://github.com/Vectorized/ERC721A-Upgradeable

We will credit you on Twitter for the next release.

iainnash:
@Vectorized Great, I'll close this out for now.

PR #201, opened 2022-03-24
Add correct function signature for _setAux

beneisnr:
Currently, the docs section for `_setAux` displays the function signature for `_getAux`. This PR includes a small update to display the correct function signature as specified in the ERC721A contract.


Vectorized:
Nice catch! Thanks!

PR #205, opened 2022-03-27
Add ERC721APausable as equivalent to ERC721Pausable

johnnyshankman:
## This PR

Adds `contracts/extensions/ERC721APausable.sol` as an option for giving developers an easy way to swap out their ERC721Pausable inherited contract implementations with something very similar in usage.

## Reference

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Pausable.sol

## Notes

Can be used alongside `ERC721ABurnable` without issues!


contracts/extensions/ERC721APausable.sol
```diff
@@ -0,0 +1,30 @@
+import "../ERC721A.sol";
+import "@openzeppelin/contracts/security/Pausable.sol";
+
```
ahbanavi, line 3:
```suggestion
// SPDX-License-Identifier: MIT
// Creator: Chiru Labs

pragma solidity ^0.8.4;

import '../ERC721A.sol';
import "@openzeppelin/contracts/security/Pausable.sol";

error ContractPaused();

```

contracts/extensions/ERC721APausable.sol
```diff
@@ -0,0 +1,30 @@
...
+     * - the contract must not be paused.
+     */
+    function _beforeTokenTransfers(
+          address from,
+          address to,
+          uint256 startTokenId,
+          uint256 quantity
+      ) internal override virtual {
+      super._beforeTokenTransfers(from, to, startTokenId, quantity);
+      require(!paused(), "Contract is paused");
```
ahbanavi, line 28:
```suggestion
      if (paused()) revert ContractPaused();
```

ahbanavi (changes requested):
LGTM,  
IMO we should use if statement with [custom error](https://blog.soliditylang.org/2021/04/21/custom-errors/) instead of `require`. See #73

contracts/extensions/ERC721APausable.sol
johnnyshankman: Surely!

contracts/extensions/ERC721APausable.sol
johnnyshankman: On it!! Great suggestion

johnnyshankman:
@ahbanavi cheers, making those updates this morning.

johnnyshankman:
Adding unit tests now!

contracts/extensions/ERC721APausable.sol
```diff
@@ -0,0 +1,37 @@
...
+     * Requirements:
+     *
+     * - the contract must not be paused.
+     */
+    function _beforeTokenTransfers(
+          address from,
+          address to,
+          uint256 startTokenId,
+          uint256 quantity
+      ) internal override virtual {
```
Vectorized, line 33: Fix indentation to 4 spaces.

johnnyshankman: Fixed! Just ran the linter. Looks like we hadn't ran it in a bit.

johnnyshankman:
@Vectorized fixed my formatting issues with the linter and picked up a few other fixes along the way. @ahbanavi wrote up those unit tests and added in an extra one for checking that it composes with `ERC721ABurnable` correctly and pauses burns.

contracts/ERC721A.sol
```diff
@@ -326,8 +326,7 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
      * Tokens start existing when they are minted (`_mint`),
      */
     function _exists(uint256 tokenId) internal view returns (bool) {
-        return _startTokenId() <= tokenId && tokenId < _currentIndex &&
-            !_ownerships[tokenId].burned;
+        return _startTokenId() <= tokenId && tokenId < _currentIndex && !_ownerships[tokenId].burned;
```
johnnyshankman, line 329: An example of a minor update the linter did on its own.

Vectorized: No prob. 👍

test/extensions/ERC721APausable.test.js
```diff
@@ -0,0 +1,73 @@
...
+        await expect(query).to.not.be.revertedWith('ContractPaused');
+        await expect(this.erc721aPausable.ownerOf(this.existingTokenID))
+          .to.be.revertedWith('OwnerQueryForNonexistentToken');
+      });
+    });
+  };
+
+describe(
+  'ERC721APausable',
+  createTestSuite({ contract: 'ERC721APausableMock', constructorArgs: ['Radioactive Pets', 'RPETS'] })
```
Vectorized, line 72: Please avoid easter eggs. 

Otherwise, it will serve as a precedent for future contributors to add in more easter eggs.

johnnyshankman: Seems harmless as this already is an easter egg and I wrote the extension but fair enough! I'll revert this commit.

johnnyshankman: Done.

johnnyshankman:
@Vectorized thank you for all the feedback! Tests and actions are passing now. @ahbanavi lmk if there's any other updates you'd like for me to make! 😄  pleasure teamworking with y'all so far 🤝

contracts/mocks/ERC721APausableMock.sol
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    function _beforeTokenTransfers(
+        address from,
+        address to,
+        uint256 startTokenId,
+        uint256 quantity
+    ) internal virtual override(ERC721A, ERC721APausable) {
+        super._beforeTokenTransfers(from, to, startTokenId, quantity);
+    }
```
ahbanavi, line 23: Is this necessary?

ahbanavi:
> @ahbanavi why was this reverted?

I just having some experiments on my fork.

Vectorized:
@johnnyshankman We are discussing about the value add of the pausable extension.

Right now, if someone were to use `Pausable` directly, they will override the `_beforeTokenTransfers` function as such
```solidity
 function _beforeTokenTransfers(
    address from,
    address to,
    uint256 startTokenId,
    uint256 quantity
) internal whenNotPaused override {
    super._beforeTokenTransfers(from, to, startTokenId, quantity);
}
```

If they use `ERC721APausable` instead, they will still have to override the `_beforeTokenTransfers` function in the deriving class, if another extension (e.g. `ERC721ABurnable`) is used. 

```solidity
function _beforeTokenTransfers(
    address from,
    address to,
    uint256 startTokenId,
    uint256 quantity
) internal virtual override(ERC721A, ERC721APausable) {
    super._beforeTokenTransfers(from, to, startTokenId, quantity);
}
```

Otherwise, solidity will throw an error:
```
TypeError: Derived contract must override function "_beforeTokenTransfers". Two or more base classes define function with same name and parameter types.
```

You can try to play about at https://wizard.openzeppelin.com/#erc721

The merit of having a `ERC721APausable` extension thus lies in:
- Helping users avoid writing the extra lines of code to override `_beforeTokenTransfers` when they are just deriving from `ERC721APausable`.
- Providing a neater way use `Pausable`, since users won't have to remember the `whenNotPaused` modifier.

contracts/extensions/ERC721APausable.sol
```diff
@@ -0,0 +1,37 @@
...
+     * - the contract must not be paused.
+     */
+    function _beforeTokenTransfers(
+        address from,
+        address to,
+        uint256 startTokenId,
+        uint256 quantity
+    ) internal virtual override {
+        super._beforeTokenTransfers(from, to, startTokenId, quantity);
+        if (paused()) revert ContractPaused();
```
holic, line 35: ooc what's the reason for this check/revert being last instead of first?

Vectorized: Good question! I believe OZ’s implementation also has it after the super call. But I see where you are coming from.

Vectorized: Wanna make a PR to OpenZeppelin in the meantime? ;)

PR #213, opened 2022-03-29
Update ERC721ABurnable.sol

kyokosdream:
Fix the parameters for call to ERC721A internal burn function. Fixes the following error:

```
TypeError: Wrong argument count for function call: 2 arguments given but expected 1.
  --> contracts/token/extensions/ERC721ABurnable.sol:21:9:
   |
21 |         _burn(tokenId, true);
   |         ^^^^^^^^^^^^^^^^^^^^


Error HH600: Compilation failed
```

### DIFF
--- a/contracts/extensions/ERC721ABurnable.sol
+++ b/contracts/extensions/ERC721ABurnable.sol
@@ -18,6 +18,6 @@ abstract contract ERC721ABurnable is ERC721A {
      * - The caller must own `tokenId` or be an approved operator.
      */
     function burn(uint256 tokenId) public virtual {
-        _burn(tokenId, true);
+        _burn(tokenId);
     }
 }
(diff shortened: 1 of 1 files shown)


kyokosdream:
Updated with a second commit that I think is the correct way to fix the problem. Instead of removing the boolean parameter from the internal _burn call, add it to the overrided function declaration.

kyokosdream:
Second commit still results in failure. Only the first proposed change works.

Vectorized:
I don't think we should remove overload and rename to `_safeBurn`.

The word **_safe_** has been used for a very different meaning in `_safeMint`.

Can you post your repo (contact me on twitter if its private)?

kyokosdream:
@Vectorized Thanks for the assistance. Closing pull request, everything working. See #214

PR #114, opened 2022-02-17
Add ERC721ALowCap + tests

Austinhs:
Ended up creating a function "tokensOfOwner" as its something Divine Anarchy uses a lot within its dApp experience. Figured I would create a pull request incase it was something you guys wanted to add to the main repo :)

PR #224, opened 2022-04-02
Add ERC721AQueryable

Vectorized:
This is made to continue from #114. 

The main concern with adding a `tokensOfOwner(owner)` function is that if the collection grows too big, the method can fail for off-chain calls (at the moment of writing, Infura has a limit of 30 times of block gas limit).

This PR adds a `tokensOfOwnerIn(owner, start, stop)` function to allow querying token IDs in a certain range. This effectively allows all token IDs to be queried, _**regardless of how big the collection gets**_. 

You can make a loop to call the function on your frontend / backend, as long as it stays below Infura's rate limits.

As such, the `ERC721ALowCap` extension will be generalizable to collections of all sizes, and should be renamed to `ERC721AQueries`.

Additional functions include `rawOwnershipOf` and `rawOwnershipsOf`, which are useful for various purposes
(imo, if the queries extension can only query `tokensOfOwner`, it feels kind of empty). 

I've used different function names for these public/external functions instead of function overloading.   
This makes it easier to use in web3 calls.

I highly recommend this, so that people won't need to copy and paste code (which can introduce errors).


ahbanavi:
WDYT about adding a `with no minted tokens` context in tests?  
To be sure that the for loops don't introduce bugs if so.

Austinhs:
I think it's important to have
```
                // All tokens have been found, we don't need to keep searching
                if (tokenIdsIdx == holdingAmount) {
                    break;
                }
```
For whatever odd reason if someone we're to call this from another contract it could potentially save some gas for them.

Vectorized:
@Austinhs It is in the boolean condition of the for loops.

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,156 @@
...
+     *   - `startTimestamp` = `<Timestamp when token was burned>`
+     *   - `burned = `true`
+     *
+     * Otherwise:
+     *   - `addr` = `<Address of owner>`
+     *   - `startTimestamp` = `<Timestamp of start of ownership>`
+     *   - `burned = `false`
+     */
+    function rawOwnershipOf(uint256 tokenId) public view returns (TokenOwnership memory) {
+        if (_exists(tokenId)) {
```
cygaar, line 32: Wouldn't it be better to fetch `_ownerships[tokenId]` first, store it in memory and then reading from it? Right now `_exists` does an SLOAD and then line 36 will result in a second SLOAD.

Austinhs:
@Vectorized 👍 

Excuse me 🙇

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,156 @@
...
+            uint256 stopLimit = _currentIndex;
+            // Set `start = max(start, _startTokenId())`.
+            if (start < _startTokenId()) {
+                start = _startTokenId();
+            }
+            // Set `stop = min(stop, _currentIndex)`.
+            if (stop > stopLimit) {
+                stop = stopLimit;
+            }
+            uint256 n = balanceOf(owner);
```
cygaar, line 81: Can we name this to `tokenBalance`? I don't like these one letter var names because they're not descriptive at all.

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,156 @@
...
+            // Set `n = min(balanceOf(owner), stop - start)`,
+            // to cater for cases where `balanceOf(owner)` is too big.
+            if (start < stop) {
+                uint256 d = stop - start;
+                if (d < n) {
+                    n = d;
+                }
+            } else {
+                n = 0;
```
cygaar, line 90: Can we just return early here? Or can we throw an error earlier?

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,156 @@
...
+            }
+            // Set `stop = min(stop, _currentIndex)`.
+            if (stop > stopLimit) {
+                stop = stopLimit;
+            }
+            uint256 n = balanceOf(owner);
+            // Set `n = min(balanceOf(owner), stop - start)`,
+            // to cater for cases where `balanceOf(owner)` is too big.
+            if (start < stop) {
+                uint256 d = stop - start;
```
cygaar, line 85: Same comment as above, `d` doesn't tell me anything

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,156 @@
...
+        return ownership;
+    }
+
+    /**
+     * @dev Returns an array of `TokenOwnership` structs at `tokenIds` in order.
+     * See {ERC721AQueries-rawOwnershipOf}
+     */
+    function rawOwnershipsOf(uint256[] memory tokenIds) external view returns (TokenOwnership[] memory) {
+        unchecked {
+            uint256 n = tokenIds.length;
```
cygaar, line 48: Let's avoid one-letter var names

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,156 @@
...
+            // to cater for cases where `balanceOf(owner)` is too big.
+            if (start < stop) {
+                uint256 d = stop - start;
+                if (d < n) {
+                    n = d;
+                }
+            } else {
+                n = 0;
+            }
+            uint256[] memory a = new uint256[](n);
```
cygaar, line 92: Same comment on naming

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,156 @@
...
+            // We need to call `rawOwnershipOf(start)`, 
+            // because the slot at `start` may not be initialized.
+            TokenOwnership memory ownership = rawOwnershipOf(start);
+            address currOwnershipAddr;
+            // If the starting slot exists, initialize `currOwnershipAddr`.
+            if (ownership.addr != address(0) && !ownership.burned) {
```
cygaar, line 98: What happens when this statement isn't true? Do we actually need to initialize `currOwnershipAddr`?

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,156 @@
...
+                currOwnershipAddr = ownership.addr;
+            }
+            for (uint256 i = start; i != stop && tokenIdsIdx != n; ++i) {
+                ownership = _ownerships[i];
+                if (ownership.burned) {
+                    continue;
+                }
+                if (ownership.addr != address(0)) {
```
cygaar, line 106: Can address be 0 and not burned?

cygaar (changes requested):

contracts/extensions/ERC721AQueries.sol
Vectorized: Return early.

contracts/extensions/ERC721AQueries.sol
Vectorized: ok

contracts/extensions/ERC721AQueries.sol
Vectorized: yes, if the slot is not initialized.

contracts/extensions/ERC721AQueries.sol
Vectorized: renamed to `windowSize`

contracts/extensions/ERC721AQueries.sol
Vectorized: renamed to `tokenIdsLength`

contracts/extensions/ERC721AQueries.sol
Vectorized: `n` renamed to `tokenIdsMaxLength`

contracts/extensions/ERC721AQueries.sol
Vectorized: When this statement isn't true, it means that the scanning window starts on a burned slot, which should be skipped in the loop.

We need to initialize `currOwnershipAddr`, because the scanning window can start in the middle on a batch that has uninitialized slots.

contracts/extensions/ERC721AQueries.sol
Vectorized: Ok, good suggestion.

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,158 @@
...
+                uint256 windowLength = stop - start;
+                if (windowLength < tokenIdsMaxLength) {
+                    tokenIdsMaxLength = windowLength;
+                }
+            } else {
+                tokenIdsMaxLength = 0;
+            }
+            uint256[] memory tokenIds = new uint256[](tokenIdsMaxLength);
+            if (tokenIdsMaxLength == 0) {
+                return tokenIds;
```
cygaar, line 95: Shouldn't you return an empty array here?

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,158 @@
...
+            uint256 tokenIdsMaxLength = balanceOf(owner);
+            // Set `tokenIdsMaxLength = min(balanceOf(owner), stop - start)`,
+            // to cater for cases where `balanceOf(owner)` is too big.
+            if (start < stop) {
+                uint256 windowLength = stop - start;
+                if (windowLength < tokenIdsMaxLength) {
+                    tokenIdsMaxLength = windowLength;
+                }
+            } else {
+                tokenIdsMaxLength = 0;
```
cygaar, line 91: Why don't you just return here?

contracts/extensions/ERC721AQueries.sol
Vectorized: I want to recycle line 93 to reduce the amount of opcodes. 
This function is external.

contracts/extensions/ERC721AQueries.sol
Vectorized: For lesser opcodes.

cygaar, line 91: I think it'd be better to check if start >= stop at the top of the function and throw an error if it's true. These are invalid parameters so it's better to error out than return 0 semantically

contracts/extensions/ERC721AQueries.sol
```diff
@@ -0,0 +1,158 @@
...
+     *   - `addr` = `<Address of owner before token was burned>`
+     *   - `startTimestamp` = `<Timestamp when token was burned>`
+     *   - `burned = `true`
+     *
+     * Otherwise:
+     *   - `addr` = `<Address of owner>`
+     *   - `startTimestamp` = `<Timestamp of start of ownership>`
+     *   - `burned = `false`
+     */
+    function rawOwnershipOf(uint256 tokenId) public view returns (TokenOwnership memory) {
```
cygaar, line 31: Might be better named as "explicitOwnershipOf"

contracts/extensions/ERC721AQueries.sol
Vectorized: Just note that even if we early revert at the top, the rest of the function still needs the same checks.

After `start` and `stop` are clamped, we will still need to check if `(start < stop)`, and then set `tokenIdsMaxLength = 0`, so that an empty array can be returned early.

contracts/extensions/ERC721AQueries.sol
Vectorized: ok

contracts/extensions/ERC721AQueries.sol
Austinhs: I agree with @Vectorized, Most of these features are for front-end queries anyways, it would be nice to be returned an empty array to handle instead of an error IMO

cygaar: It might be easier to handle on the frontend if you gracefully handle the error, but it doesn't mean it's the right way to do this. If you look at most of the functions in OZ's contracts, they all error out whenever you pass in invalid input parameters.

@Vectorized why do you need to keep the rest of the function the same? You will definitely have to change some tests if you make this change, I wouldn't expect the existing tests to pass.

Vectorized: If I remove the checks and set to zero after the clamping, the function can go into a near infinite loop.

cygaar: What's the exact scenario you're thinking of? I don't see why we still need the if-statement on line 88. Line 93 can never evaluate to true

Vectorized: Suppose a collection with `_currentIndex` = 8, and we pass in `start` = 10 and `stop` = 11. After clamping, `start` will remain 10, and `stop` will be 8.

Vectorized: I don’t want to add another custom error for the out of bounds scenario.

Vectorized: On custom errors, I am on the side that they are a poorly implemented feature of solidity and should be avoided as much as possible. 

If this PR is too troublesome to satisfy all parties, I suggest reverting the LowCap extension and leave it to end users to do their own implementation.

cygaar: What about doing `if (start >= stop || start >= stopLimit || stop <= _startTokenId) revert InvalidQueryRange();` after line 80? Basically, do the input validation after you know the start index and stop limit.

Vectorized: That will cause it to revert for empty collections, which is not desired.

tommypacker (approved):

cygaar (approved):

ahbanavi (approved):

Austinhs (approved):

PR #231, opened 2022-04-08
Docs update

Vectorized:
- Upgradeable docs
- ERC721AQueries docs
- ERC721APausable docs
- Various small edits

To be merged just before publishing next version.


docs/erc721a-queries.md
```diff
@@ -0,0 +1,84 @@
...
+Returns an array of token IDs owned by `owner`.
+
+This function scans the ownership mapping and is O(totalSupply) in complexity.  
+It is meant to be called off-chain.
+
+See [`tokensOfOwnerIn`](#tokensOfOwnerIn) for splitting the scan into 
+multiple smaller scans if the collection is large enough to 
+cause an out-of-gas error. 
+
+This function should work fine for 10k PFP collections.
```
Austinhs, line 67: Thoughts on maybe adding a similar warning for a size that would not work?

Austinhs: Something like:

Should be avoided for collections > 30K (arbitrary number, btw), and should NOT be used for dynamic supplies like breeding.

Vectorized: I am purposely vague about the limit, because different RPC providers may have different gas limits on view functions.

Austinhs (approved):

PR #232, opened 2022-04-08
Add Update Version Script

### LINKED ISSUE #138: Suggestion: add contract version
It would be a good idea to add the release version on top of the [contracts/ERC721A.sol](https://github.com/chiru-labs/ERC721A/blob/main/contracts/ERC721A.sol) file and update it when a new version releases.  
Some devs or savvy people would like to know what version of `ERC721A` the implemented contract uses. 

```solidity
// SPDX-License-Identifier: MIT
// ERC721A v2.2.0
// Creator: Chiru Labs
.
.
.
```

ahbanavi:
Fixes #138, If the maintainers consider using this script before tagging a new version.

More info: https://github.com/chiru-labs/ERC721A/issues/138#issuecomment-1091208853


scripts/release/update-version.js
```diff
@@ -0,0 +1,39 @@
...
+    // read file content
+    const content = fs.readFileSync(file, 'utf8');
+
+    const versionStringLine = versionPrefix + version;
+
+    let updatedContent;
+    if (content.includes(versionPrefix)) {
+      updatedContent = content.replace(new RegExp(`${versionPrefix}.*`), `${versionStringLine}`);
+    } else {
+      updatedContent = content.replace(new RegExp(`${spdxString}`), `${spdxString}\n${versionStringLine}`);
```
Austinhs, line 33: Why is this in the else? Would'nt we always want to change the `spdxString` incase this script is used to update the License as well?

Vectorized:
Just a heads up, we may postpone this to the release after upgradeable.

scripts/release/update-version.js
ahbanavi: This script doesn't ment to update the License. The usage of `spdxString` is for the contracts that don't have a version line.  

Suppose a contract already has a version line. In that case, we are just updating that line, but some contracts may be added later and don't have a version string; for those, we are using `spdxString` and adding the version line after it.

PR #235, opened 2022-04-12
Add _mint(to, quantity) function

Vectorized:
Most people use `_mint(to, quantity, "", false)` when they do not need safe minting.

Sadly, the compiler will still produce the bytecode needed for safe minting.

_\<Inserts optimizoor meme here\>_

This PR adds a `_mint(to, quantity)` function.

It saves 100 gas per call, and 42495 gas during deployment (about 200 bytes), which is about 6% for a minimal contract.

The code is manually inlined directly. Simply calling `_mint(to, quantity, "", false)` inside actually leads to increased gas costs (50 gas function call overhead) with no reduction in bytecode size. 

⛽ Gas Reports

<details><summary>Before</summary>
<pre>
·---------------------------------------|---------------------------|-------------|
|         Solc version: 0.8.11          ·  Optimizer enabled: true  ·  Runs: 800  ·
········································|···························|·············|
|  Methods                                                                         
····························|···········|·············|·············|·············|
|  Contract                 ·  Method   ·  Min        ·  Max        ·  Avg        ·
····························|···········|·············|·············|·············|
|  ERC721AGasReporterMock2  ·  mintOne  ·      56462  ·      90662  ·      57146  ·
····························|···········|·············|·············|·············|
|  ERC721AGasReporterMock2  ·  mintTen  ·      74116  ·     108316  ·      74800  ·
····························|···········|·············|·············|·············|
|  Deployments                          ·                                         ·
········································|·············|·············|·············|
|  ERC721AGasReporterMock2              ·          -  ·          -  ·    1128702  ·
·---------------------------------------|-------------|-------------|-------------|
</pre> 
</details>

<details><summary>After</summary>
<pre>
·---------------------------------------|---------------------------|-------------|
|         Solc version: 0.8.11          ·  Optimizer enabled: true  ·  Runs: 800  ·
········································|···························|·············|
|  Methods                                                                         
····························|···········|·············|·············|·············|
|  Contract                 ·  Method   ·  Min        ·  Max        ·  Avg        ·
····························|···········|·············|·············|·············|
|  ERC721AGasReporterMock2  ·  mintOne  ·      56361  ·      90561  ·      57045  ·
····························|···········|·············|·············|·············|
|  ERC721AGasReporterMock2  ·  mintTen  ·      74015  ·     108215  ·      74699  ·
····························|···········|·············|·············|·············|
|  Deployments                          ·                                         ·
········································|·············|·············|·············|
|  ERC721AGasReporterMock2              ·          -  ·          -  ·    1086207  ·
·---------------------------------------|-------------|-------------|-------------|
</pre>
</details>

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -406,6 +406,38 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
     }
 
     /**
+     * @dev Equivalent to _mint(to, quantity, '', false).
+     */
+    function _mint(address to, uint256 quantity) internal {
+        uint256 startTokenId = _currentIndex;
+        if (to == address(0)) revert MintToZeroAddress();
+        if (quantity == 0) revert MintZeroQuantity();
+
+        _beforeTokenTransfers(address(0), to, startTokenId, quantity);
+
+        // Overflows are incredibly unrealistic.
+        // balance or numberMinted overflow if current value of either + quantity > 1.8e19 (2**64) - 1
+        // updatedIndex overflows if _currentIndex + quantity > 1.2e77 (2**256) - 1
+        unchecked {
+            _addressData[to].balance += uint64(quantity);
+            _addressData[to].numberMinted += uint64(quantity);
+
+            _ownerships[startTokenId].addr = to;
+            _ownerships[startTokenId].startTimestamp = uint64(block.timestamp);
+
+            uint256 updatedIndex = startTokenId;
+            uint256 end = updatedIndex + quantity;
+
+            do {
+                emit Transfer(address(0), to, updatedIndex++);
+            } while (updatedIndex != end);
+
+            _currentIndex = updatedIndex;
+        }
+        _afterTokenTransfers(address(0), to, startTokenId, quantity);
+    }
+
+    /**
      * @dev Transfers `tokenId` from `from` to `to`.
      *
      * Requirements:
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -329,6 +329,9 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
         return _startTokenId() <= tokenId && tokenId < _currentIndex && !_ownerships[tokenId].burned;
     }
 
+    /**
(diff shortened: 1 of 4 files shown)


tunnckoCore (approved):

ahbanavi (approved):
Wow, this makes my life a lot easier.  
I was frustrated doing `_mint(addr, quantity, '', false)` instead of a simple `_mint(addr, quantity)` throughout my contract.  

Apart from being more user-friendly, it's more optimized too (if an implemented contract doesn't use both _mint and _safeMint).  

The downside, lower maintainability, is totally worth it. "pretends to be a maintainer"

contracts/ERC721A.sol
```diff
@@ -405,6 +405,38 @@ contract ERC721A is Context, ERC165, IERC721, IERC721Metadata {
         _afterTokenTransfers(address(0), to, startTokenId, quantity);
     }
 
+    /**
+     * @dev Equivalent to _mint(to, quantity, '', false).
+     */
+    function _mint(address to, uint256 quantity) internal {
```
cygaar, line 411: Should we just rename the existing mint function to _safeMint and remove the `safe` parameter? This will be backwards incompatible, but will make the code cleaner since I don't see ppl calling `_mint(to, qty,'', false)` going forward

ahbanavi: Agree to remove the `safe` parameter and rename `_safeMint`. 
For backward compatibility, we could add back `_mint(to, qty, data, safe)` and then call `_mint` and `_safeMint` based on the `safe` parameter, but it cost maintainability.

Vectorized: I’m just worried about the customers calling the ERC721A helpdesk if we break backwards compatibility. Lemme think for a while.

cygaar: Well we've broken it a couple times already. I'm ok cutting a v4 of ERC721A for a change like that

Vectorized: Outside rn, brb in a few hours probably

Vectorized: Code is neater, about 50 gas saved. Let’s go.

PR #251, opened 2022-04-22
Changing functions to external

zestrells:
Scanned ERC721A with some static analysis tools and updating based on findings.


Vectorized:
These functions are intentionally left public so that they can be called from within derived contracts.

zestrells:
> These functions are intentionally left public so that they can be called from within derived contracts.

Doesn't the external declaration still suffice and leave it public?

Vectorized:
https://docs.soliditylang.org/en/v0.4.24/contracts.html#visibility-and-getters  

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol

PR #252, opened 2022-04-25
Included Upgradeable Contract

0xAlfa:
This is the approach taken by 0xytocin Labs for making the ERC721A contract upgradeable.


Vectorized:
https://github.com/chiru-labs/ERC721A-Upgradeable 

We already have upgradeable contracts.

PR #255, opened 2022-04-28
add override and remove immutable in function to fit version under v0.8.8

tn606024:
contract's solidity version is solidity ^0.8.4, but in code use two feature only support ^0.8.8.

https://blog.soliditylang.org/2021/09/27/solidity-0.8.8-release-announcement/

you can see **No Override for Interface Functions** and **Reading From Immutable Variables** in this article.

so to fit contract version under 0.8.8, I add override for Interface function and remove read immutable variable in one mock contract.


Vectorized:
Thanks! Nice catch.

PR #257, opened 2022-04-29
Remove Pausable

Vectorized:


holic:
why was this removed, ooc?

PR #269, opened 2022-05-08
Add ERC721ALockable

maplerichie:
Inspire by the guardian concept.
Lockable token allow a token remain in hot wallet for ease of use like sign for ownership proof while locked with a cold wallet. And prevent from steal of accidentally token approval.


Vectorized:
Thank you for the great efforts.

In my opinion, this should be in a separate library that can work with multiple ERC721 implementations (OpenZeppelin, Solmate, ERC721A). 

Additionally, reverting in the `_beforeTokensTransfer` function can pose UX issues: people can list locked tokens below the floor price to cause FUD. 

Ideally, this should be formalized into an EIP so that marketplaces know how to handle tokens that revert before transfer. 

Also, additional documentation and Natspec comments will have to be maintained for every extension we accept. 

We have to explain why we include certain functions in a certain way, document the potential limitations, etc.

Issue #270, opened 2022-05-08
Golfing ERC721A


0xPhaze:
Hey,

I tried bringing the costs down for ERC721A. Here's a gas report with a small readme:
https://github.com/0xPhaze/ERC721A-Improved

- Internally, `tokenData` is passed as storage instead of memory.
- Keeps track of whether the next token data has already been set before, so that we don't have to check it (1 cold sload) on every transfer for all eternity.
- Saves (warm) sload on variables such as `collectionSize` and `maxPerWallet`.

| Function                      |    Gas |
| :---------------------------- | -----: |
| _mint 1_                      |        |
| test_mint1_ERC721A()          |  59713 |
| test_mint1_ERC721AX()         |  59333 |
| _mint 5_                      |        |
| test_mint5_ERC721A()          |  67486 |
| test_mint5_ERC721AX()         |  67131 |
| _transfer 1_                  |        |
| test_transferFrom1_ERC721A()  |  73837 |
| test_transferFrom1_ERC721AX() |  49435 |
| _transfer 5_                  |        |
| test_transferFrom5_ERC721A()  | 154195 |
| test_transferFrom5_ERC721AX() | 127657 |

Wanted to ask first, if ERC721A is interested having some of these optimizations implemented.

Vectorized:
Make a PR.

You may need to make new tests under GasUsage.test.js

Vectorized:
If you really want to go for it, you can even make it such that the data are all stored as uint256 internally, and only converted into structs in public / external functions. 

The next version will be a major release, so we can afford some breaking changes.

Vectorized:
I'm waiting for review on #272. Once approved, let's continue from there.

Vectorized:
@0xPhaze I've added in #272. Will credit you in the next release. Big thanks!

0xPhaze:
Awesome, thanks!
Although, I think that doesn't include the 'nextTokenDataSet' boolean which is a big part of the optimization. Instead of checking whether the next packed ownership has been set already by loading its storage, this would keep track of whether it has already previously been set and save that sload for all future transfers.

Issue #267, opened 2022-05-06
Use Diamond Storage


mudgen:
It would be useful if the ERC721A implementation was changed to use [diamond storage](https://dev.to/mudgen/how-diamond-storage-works-90e).

Diamond storage is a contract storage technique that makes organizing and upgrading data in proxy contracts and [diamonds](https://eips.ethereum.org/EIPS/eip-2535) possible.

If the ERC721A implementation used diamond storage then it could be used in diamonds and other proxy contracts without change.

cliffhall:
This is true, and one of my biggest complaints about most of the 721 implementations, they cannot be used with a diamond if they use the common storage approach. Even OZ's upgradeable suite does this. :(

Vectorized:
https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2964

mudgen:
> [OpenZeppelin/openzeppelin-contracts#2964](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2964)

Yes, OpenZeppelin plans to adopt diamond storage, but we don't know when that will be:

> From [frangio](https://github.com/frangio): Yes, we want to adopt diamond storage for the next major release, but we don't have an estimated release date. It will most likely happen whenever Solidity 0.9 is released, and as far as we know there is no date for that either.

PR #272, opened 2022-05-09
Use uint packing instead of structs

Vectorized:
This PR is made in preparation for #267. 

This is a breaking change, as the internal `_ownerships` mapping is renamed to `_packedOwnerships`.

Using uints allow us to write most of the packing logic in Solidity instead of assembly. This allows us to keep the code compatible with the diamonds transpiler we are planning to use in the near future. Also, it is easier to read, and surprisingly more performant. 

This PR is inspired by and includes #270. I’ve tried the idea with structs: it added quite a bit of overhead because of the many values that are often needlessly unpacked by the compiler. Manual unpacking allows us to achieve a much lower overhead. 

Additional tests have been added to ensure that the `startTimestamp`s are packed / unpacked properly. 

**Before:**

```
·---------------------------------------------------------------|---------------------------|-------------|
|                     Solc version: 0.8.11                      ·  Optimizer enabled: true  ·  Runs: 800  ·
································································|···························|·············|
|  Methods                                                                                                 
··········································|·····················|·············|·············|·············|
|  Contract                               ·  Method             ·  Min        ·  Max        ·  Avg        ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  approve            ·      30777  ·      50917  ·      40847  ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  burn               ·      46142  ·     112535  ·      92768  ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  safeMint           ·          -  ·          -  ·     111106  ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  setApprovalForAll  ·      26277  ·      46189  ·      36233  ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  transferFrom       ·      84577  ·      90863  ·      87720  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  mintOne            ·      56358  ·      90558  ·      57042  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  mintTen            ·      73985  ·     108185  ·      75629  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  safeMintOne        ·      59094  ·      93294  ·      59778  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  safeMintTen        ·      76699  ·     110899  ·      77383  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  transferFrom       ·      47175  ·      86955  ·      49164  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  approve            ·          -  ·          -  ·      50849  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  burn               ·          -  ·          -  ·      63346  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  mint               ·      90667  ·      98499  ·      96413  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  safeMint           ·      78306  ·     123064  ·      87877  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  safeTransferFrom   ·      91614  ·      91626  ·      91623  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  setApprovalForAll  ·      46195  ·      46219  ·      46216  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  setAux             ·      27144  ·      44328  ·      32872  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  transferFrom       ·      84575  ·      84599  ·      84594  ·
··········································|·····················|·············|·············|·············|
|  Deployments                                                  ·                                         ·
································································|·············|·············|·············|
|  ERC721ABurnableMock                                          ·          -  ·          -  ·    1300089  ·
································································|·············|·············|·············|
|  ERC721AGasReporterMock                                       ·          -  ·          -  ·    1213358  ·
································································|·············|·············|·············|
|  ERC721AMock                                                  ·          -  ·          -  ·    1464231  ·
·---------------------------------------------------------------|-------------|-------------|-------------|
```

**After:**

```
·---------------------------------------------------------------|---------------------------|-------------|
|                     Solc version: 0.8.11                      ·  Optimizer enabled: true  ·  Runs: 800  ·
································································|···························|·············|
|  Methods                                                                                                 
··········································|·····················|·············|·············|·············|
|  Contract                               ·  Method             ·  Min        ·  Max        ·  Avg        ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  approve            ·      30453  ·      50593  ·      40523  ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  burn               ·      43393  ·     111032  ·      91192  ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  safeMint           ·          -  ·          -  ·     110959  ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  setApprovalForAll  ·      26277  ·      46189  ·      36233  ·
··········································|·····················|·············|·············|·············|
|  ERC721ABurnableMock                    ·  transferFrom       ·      83946  ·      89685  ·      86816  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  mintOne            ·      56217  ·      90417  ·      56901  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  mintTen            ·      73844  ·     108044  ·      75488  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  safeMintOne        ·      58947  ·      93147  ·      59631  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  safeMintTen        ·      76552  ·     110752  ·      77236  ·
··········································|·····················|·············|·············|·············|
|  ERC721AGasReporterMock                 ·  transferFrom       ·      44489  ·      86180  ·      46574  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  approve            ·          -  ·          -  ·      50547  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  burn               ·          -  ·          -  ·      60597  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  mint               ·      90526  ·      98358  ·      96272  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  safeMint           ·      78159  ·     122917  ·      87730  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  safeTransferFrom   ·      91008  ·      91020  ·      91017  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  setApprovalForAll  ·      46195  ·      46219  ·      46216  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  setAux             ·      27145  ·      44329  ·      32873  ·
··········································|·····················|·············|·············|·············|
|  ERC721AMock                            ·  transferFrom       ·      83966  ·      83990  ·      83985  ·
··········································|·····················|·············|·············|·············|
|  Deployments                                                  ·                                         ·
································································|·············|·············|·············|
|  ERC721ABurnableMock                                          ·          -  ·          -  ·    1213610  ·
································································|·············|·············|·············|
|  ERC721AGasReporterMock                                       ·          -  ·          -  ·    1116449  ·
································································|·············|·············|·············|
|  ERC721AMock                                                  ·          -  ·          -  ·    1329475  ·
·---------------------------------------------------------------|-------------|-------------|-------------|
```

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -292,7 +292,7 @@ contract ERC721A is IERC721A {
         address owner = address(uint160(_packedOwnershipOf(tokenId)));
         if (to == owner) revert ApprovalToCurrentOwner();
 
-        if (_msgSenderERC721A() != owner) if(!isApprovedForAll(owner, _msgSenderERC721A())) {
+        if (_msgSenderERC721A() != owner) if (!isApprovedForAll(owner, _msgSenderERC721A())) {
             revert ApprovalCallerNotOwnerNorApproved();
         }
 
@@ -623,6 +623,7 @@ contract ERC721A is IERC721A {
             // - `numberBurned += 1`.
             //
             // We can directly decrement the balance, and increment the number burned.
+            // This is equivalent to `packed -= 1; packed += 1 << BITPOS_NUMBER_BURNED;`.
             _packedAddressData[from] += (1 << BITPOS_NUMBER_BURNED) - 1;
 
             { // Scoped for extra gas optimization.
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -403,25 +403,6 @@ contract ERC721A is IERC721A {
     }
 
     /**
-     * @dev Writes the ownership data for mint.
-     */
-    function _mintWriteOwnership(
-        address to,
-        uint256 quantity,
-        uint256 startTokenId
-    ) private {
-        // Updates:
-        // - `address` to the owner.
-        // - `startTimestamp` to the timestamp of minting.
-        // - `burned` to `false`.
-        // - `nextInitialized` to `quantity == 1`.
-        _packedOwnerships[startTokenId] =
-            _addressToUint256(to) |
-            (block.timestamp << BITPOS_START_TIMESTAMP) |
-            (_boolToUint256(quantity == 1) << BITPOS_NEXT_INITIALIZED);
-    }
-
-    /**
      * @dev Safely mints `quantity` tokens and transfers them to `to`.
      *
      * Requirements:
@@ -454,7 +435,15 @@ contract ERC721A is IERC721A {
             // We can directly add to the balance and number minted.
             _packedAddressData[to] += quantity | (quantity << BITPOS_NUMBER_MINTED);
 
-            _mintWriteOwnership(to, quantity, startTokenId);
+            // Updates:
+            // - `address` to the owner.
+            // - `startTimestamp` to the timestamp of minting.
+            // - `burned` to `false`.
+            // - `nextInitialized` to `quantity == 1`.
+            _packedOwnerships[startTokenId] =
+                _addressToUint256(to) |
+                (block.timestamp << BITPOS_START_TIMESTAMP) |
+                (_boolToUint256(quantity == 1) << BITPOS_NEXT_INITIALIZED);
 
             uint256 updatedIndex = startTokenId;
             uint256 end = updatedIndex + quantity;
@@ -506,7 +495,15 @@ contract ERC721A is IERC721A {
             // We can directly add to the balance and number minted.
             _packedAddressData[to] += quantity | (quantity << BITPOS_NUMBER_MINTED);
 
-            _mintWriteOwnership(to, quantity, startTokenId);
+            // Updates:
+            // - `address` to the owner.
+            // - `startTimestamp` to the timestamp of minting.
+            // - `burned` to `false`.
+            // - `nextInitialized` to `quantity == 1`.
+            _packedOwnerships[startTokenId] =
+                _addressToUint256(to) |
+                (block.timestamp << BITPOS_START_TIMESTAMP) |
+                (_boolToUint256(quantity == 1) << BITPOS_NEXT_INITIALIZED);
 
             uint256 updatedIndex = startTokenId;
             uint256 end = updatedIndex + quantity;
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -42,11 +42,14 @@ contract ERC721A is IERC721A {
     string private _symbol;
 
     // Mapping from token ID to ownership details
-    // An empty struct value does not necessarily mean the token is unowned. See _ownershipOf implementation for details.
-    mapping(uint256 => TokenOwnership) internal _ownerships;
+    // An empty struct value does not necessarily mean the token is unowned. 
+    // See _ownershipOf implementation for details.
+    // (addr) | (startTimestamp << 160) | (burned << 224) | (nextTokenDataSet << 225) 
+    mapping(uint256 => uint256) internal _packedOwnerships;
 
-    // Mapping owner address to address data
-    mapping(address => AddressData) private _addressData;
+    // Mapping owner address to address data.
+    // (balance) | (numberMinted << 64) | (numberBurned << 128) | (aux << 192) 
+    mapping(address => uint256) private _packedAddressData;
 
     // Mapping from token ID to approved address
     mapping(uint256 => address) private _tokenApprovals;
@@ -107,28 +110,28 @@ contract ERC721A is IERC721A {
      */
     function balanceOf(address owner) public view override returns (uint256) {
         if (owner == address(0)) revert BalanceQueryForZeroAddress();
-        return uint256(_addressData[owner].balance);
+        return _packedAddressData[owner] & ((1 << 64) - 1);
     }
 
     /**
      * Returns the number of tokens minted by `owner`.
      */
     function _numberMinted(address owner) internal view returns (uint256) {
-        return uint256(_addressData[owner].numberMinted);
+        return (_packedAddressData[owner] >> 64) & ((1 << 64) - 1);
     }
 
     /**
      * Returns the number of tokens burned by or on behalf of `owner`.
      */
     function _numberBurned(address owner) internal view returns (uint256) {
-        return uint256(_addressData[owner].numberBurned);
+        return (_packedAddressData[owner] >> 128) & ((1 << 64) - 1);
     }
 
     /**
      * Returns the auxillary data for `owner`. (e.g. number of whitelist mint slots used).
      */
     function _getAux(address owner) internal view returns (uint64) {
-        return _addressData[owner].aux;
+        return uint64(_packedAddressData[owner] >> 192);
     }
 
     /**
@@ -136,22 +139,23 @@ contract ERC721A is IERC721A {
      * If there are multiple variables, please pack them into a uint64.
      */
     function _setAux(address owner, uint64 aux) internal {
-        _addressData[owner].aux = aux;
+        uint256 packed = _packedAddressData[owner];
+        packed = (packed & ((1 << 192) - 1)) | (uint256(aux) << 192);
+        _packedAddressData[owner] = packed;
     }
 
-    /**
-     * Gas spent here starts off proportional to the maximum mint batch size.
-     * It gradually moves to O(1) as tokens get transferred around in the collection over time.
-     */
-    function _ownershipOf(uint256 tokenId) internal view returns (TokenOwnership memory) {
+    function _packedOwnershipOf(uint256 tokenId) private view returns (uint256) {
         uint256 curr = tokenId;
 
         unchecked {
             if (_startTokenId() <= curr) if (curr < _currentIndex) {
-                TokenOwnership memory ownership = _ownerships[curr];
-                if (!ownership.burned) {
-                    if (ownership.addr != address(0)) {
-                        return ownership;
+                uint256 packed = _packedOwnerships[curr];
+                uint256 burned = packed & (1 << 224);
+                uint256 addr;
+                if (burned == 0) {
+                    addr = packed & ((1 << 160) - 1);
+                    if (addr != 0) {
+                        return packed;
                     }
                     // Invariant:
                     // There will always be an ownership that has an address and is not burned
@@ -159,15 +163,40 @@ contract ERC721A is IERC721A {
                     // Hence, curr will not underflow.
                     while (true) {
                         curr--;
-                        ownership = _ownerships[curr];
-                        if (ownership.addr != address(0)) {
-                            return ownership;
+                        packed = _packedOwnerships[curr];
+                        addr = packed & ((1 << 160) - 1);
+                        if (addr != 0) {
+                            return packed;
                         }
                     }
                 }
             }
         }
         revert OwnerQueryForNonexistentToken();
+    }
+
+    /**
+     * Returns the unpacked `TokenOwnership` struct from `packed`.
+     */
+    function _unpackedOwnership(uint256 packed) private pure returns (TokenOwnership memory ownership) {
+        ownership.addr = address(uint160(packed));
+        ownership.startTimestamp = uint64(packed >> 160);
+        ownership.burned = packed & (1 << 224) != 0;
+    }
+
+    /**
+     * Returns the unpacked `TokenOwnership` struct at `index`.
+     */
+    function _ownershipAt(uint256 index) internal view returns (TokenOwnership memory) {
+        return _unpackedOwnership(_packedOwnerships[index]);
+    }
+
+    /**
+     * Gas spent here starts off proportional to the maximum mint batch size.
+     * It gradually moves to O(1) as tokens get transferred around in the collection over time.
+     */
+    function _ownershipOf(uint256 tokenId) internal view returns (TokenOwnership memory) {
+        return _unpackedOwnership(_packedOwnershipOf(tokenId));
     }
 
     /**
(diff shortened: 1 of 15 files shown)


contracts/ERC721A.sol
```diff
@@ -324,11 +356,17 @@ contract ERC721A is Context, ERC165, IERC721A {
...
-            _ownerships[startTokenId].addr = to;
-            _ownerships[startTokenId].startTimestamp = uint64(block.timestamp);
+            { // Scoped for extra gas optimization.
+                uint256 packed;
+                assembly {
+                    // uint256(uint160(to)) | (block.timestamp << 160)
+                    packed := or(to, shl(160, timestamp()))
+                }
+                _packedOwnerships[startTokenId] = packed;    
+            }
```
Vectorized, line 369: This is to prepare for a branchless implementation of #270 
```solidity
assembly {
    // uint256(uint160(to)) | (block.timestamp << 160) | ((quantity == 1 ? 1 : 0) << 225)
    packed := or(or(to, shl(160, timestamp())), shl(225, eq(quantity, 1)))
}

**Comment by 0xPhaze** [comment]
~ just an idea:
Accessing and writing to the packed ownership could be handled by a library to make things more readable.
I had done something like that here in this repo:  https://github.com/0xPhaze/ERC721M/blob/master/src/ERC721MMC.sol
I even explored (but never implemented, because it was a lot of wrap/unwrap) using custom types for a packed uint256 to further guide access to its packed storage. This could be interesting for ERC721A.

**Comment by Vectorized** [comment]
I think it's still okay to directly write the bitwise ops inline to have a little more control.

File: contracts/ERC721A.sol [hunk 2]
```diff
@@ -107,74 +120,106 @@ contract ERC721A is IERC721A {
...
-        return uint256(_addressData[owner].balance);
+        return _packedAddressData[owner] & ((1 << 64) - 1);
     }
 
     /**
      * Returns the number of tokens minted by `owner`.
      */
     function _numberMinted(address owner) internal view returns (uint256) {
-        return uint256(_addressData[owner].numberMinted);
+        return (_packedAddressData[owner] >> 64) & ((1 << 64) - 1);
```
cygaar, line 130: I think we can make these bit masks constants so the shift operation only needs to be done once. It also gives us some clearer naming

contracts/ERC721A.sol
```diff
@@ -107,74 +120,106 @@ contract ERC721A is IERC721A {
...
     }
 
     /**
      * Sets the auxillary data for `owner`. (e.g. number of whitelist mint slots used).
      * If there are multiple variables, please pack them into a uint64.
      */
     function _setAux(address owner, uint64 aux) internal {
-        _addressData[owner].aux = aux;
+        uint256 packed = _packedAddressData[owner];
+        packed = (packed & ((1 << 192) - 1)) | (uint256(aux) << 192);
```
cygaar, line 153: Same comment about making the masks constants

contracts/ERC721A.sol
```diff
@@ -296,7 +341,8 @@ contract ERC721A is IERC721A {
      * Tokens start existing when they are minted (`_mint`),
      */
     function _exists(uint256 tokenId) internal view returns (bool) {
-        return _startTokenId() <= tokenId && tokenId < _currentIndex && !_ownerships[tokenId].burned;
+        return _startTokenId() <= tokenId && tokenId < _currentIndex && // If within bounds,
+            _packedOwnerships[tokenId] & (1 << 224) == 0; // and not burned.
```
cygaar, line 345: It might be more readable is we extract `_packedOwnerships[tokenId] & (1 << 224)` into an internal helper function called getBurned or something of that nature. Also makes it easier for inheriting contracts to check if a token is burned

cygaar: Could probably do this for all the fields we used to have on the ownership and address structs

contracts/ERC721A.sol
```diff
@@ -332,11 +378,30 @@ contract ERC721A is IERC721A {
...
-            _ownerships[startTokenId].addr = to;
-            _ownerships[startTokenId].startTimestamp = uint64(block.timestamp);
+            // Updates:
+            // - `balance += quantity`.
+            // - `numberMinted += quantity`.
+            //
+            // We can directly add to the balance and number minted.
+            _packedAddressData[to] += quantity | (quantity << 64);
+
+            { // Scoped for extra gas optimization.
```
cygaar, line 388: How much gas does scoping save here?

contracts/ERC721A.sol
```diff
@@ -381,11 +446,30 @@ contract ERC721A is IERC721A {
...
+                uint256 packed;
```
cygaar, line 457: Can we move this block of code into a helper function and reference it here and in _safeMint so we can reduce code duplication?

contracts/ERC721A.sol
```diff
@@ -498,26 +595,39 @@ contract ERC721A is IERC721A {
...
-                // as a burned slot cannot contain the zero address.
-                if (nextTokenId != _currentIndex) {
-                    nextSlot.addr = from;
-                    nextSlot.startTimestamp = prevOwnership.startTimestamp;
+            // Updates:
+            // - `balance -= 1`.
+            // - `numberBurned += 1`.
+            //
+            // We can directly decrement the balance, and increment the number burned.
+            _packedAddressData[from] += (1 << 128) - 1;
```
cygaar, line 603: How does this update balance and numberBurned at the same time?

cygaar (changes requested):

contracts/ERC721A.sol
Vectorized: The compiler will evaluate the constant expressions on compile time.

contracts/ERC721A.sol
Vectorized: Inheriting contracts won't be able to read and write the raw bits.

contracts/ERC721A.sol
Vectorized: 11 gas.

contracts/ERC721A.sol
Vectorized: Ok, fortunately, the compiler will inline them properly.

contracts/ERC721A.sol
Vectorized: It subtracts 1 from bits [0..64], and adds 1 to bits [128..192].

contracts/ERC721A.sol
cygaar: Even so, I prefer not to have magic numbers in our code if possible

contracts/ERC721A.sol
cygaar: Are you saying they shouldn't be able to or they won't be able to?

Vectorized: They shouldn't and won't be able to.

contracts/ERC721A.sol
Vectorized: Ok. The code will be more verbose though.

contracts/ERC721A.sol
```diff
@@ -29,6 +29,36 @@ interface ERC721A__IERC721Receiver {
...
+    uint256 private constant BITPOS_BURNED = 224;
+
+    // The bit position of the `nextInitialized` bit in packed ownership.
+    uint256 private constant BITPOS_NEXT_INITIALIZED = 225;
+
+    // The bit mask of the `burned` bit in packed ownership.
+    uint256 private constant BITMASK_BURNED = 1 << 224;
+
+    // The bit mask of the `nextInitialized` bit in packed ownership.
+    uint256 private constant BITMASK_NEXT_INITIALIZED = 1 << 225;
```
Vectorized, line 60: Note that the expression cannot refer to other constants (e.g. `1 << BITPOS_NEXT_INITIALIZED`), or else the constants cannot be used in assembly. 

Compiler limitation.

contracts/ERC721A.sol
```diff
@@ -107,74 +150,106 @@ contract ERC721A is IERC721A {
...
         revert OwnerQueryForNonexistentToken();
     }
 
+    /**
+     * Returns the unpacked `TokenOwnership` struct from `packed`.
+     */
+    function _unpackedOwnership(uint256 packed) private pure returns (TokenOwnership memory ownership) {
+        ownership.addr = address(uint160(packed));
+        ownership.startTimestamp = uint64(packed >> BITPOS_START_TIMESTAMP);
+        ownership.burned = packed & BITMASK_BURNED != 0;
```
Vectorized, line 221: Directly masking will be better than including a function. 

Returning boolean can have overhead.

test/extensions/ERC721ABurnable.test.js
```diff
@@ -106,6 +112,22 @@ const createTestSuite = ({ contract, constructorArgs }) =>
...
+        const timestampToMine = (await getBlockTimestamp()) + 100;
+        await mineBlockTimestamp(timestampToMine);
+        const timestampMined = await getBlockTimestamp();
+        await this.erc721aBurnable.connect(this.addr1).burn(tokenIdToBurn);
+        const ownershipAfter = await this.erc721aBurnable.getOwnershipAt(tokenIdToBurn);
+        const timestampAfter = parseInt(ownershipAfter.startTimestamp);
+        expect(timestampBefore).to.be.lt(timestampToMine);
+        expect(timestampAfter).to.be.gte(timestampToMine);
+        expect(timestampAfter).to.be.gt(timestampBefore);
+        expect(timestampToMine).to.be.eq(timestampMined);
```
Vectorized, line 128: Unlike Foundry, the next transaction after mining a timestamp may have a `block.timestamp` that is slightly higher than the mined timestamp.

Which is why we need these many checks.

contracts/ERC721A.sol
Vectorized: Added a comment to make it clearer.

contracts/ERC721A.sol
Vectorized, line 60: The constant expressions here must not include other constants (e.g. `1 << BITPOS_NEXT_INITIALIZED`), otherwise, the compiler may not fully evaluate them on compile time, causing overhead.

contracts/ERC721A.sol
```diff
@@ -332,11 +450,14 @@ contract ERC721A is IERC721A {
...
         // updatedIndex overflows if _currentIndex + quantity > 1.2e77 (2**256) - 1
         unchecked {
-            _addressData[to].balance += uint64(quantity);
-            _addressData[to].numberMinted += uint64(quantity);
+            // Updates:
+            // - `balance += quantity`.
+            // - `numberMinted += quantity`.
+            //
+            // We can directly add to the balance and number minted.
+            _packedAddressData[to] += quantity | (quantity << BITPOS_NUMBER_MINTED);
```
Vectorized, line 458: Moving this into a function will cause overhead.

contracts/ERC721A.sol
```diff
@@ -434,23 +558,30 @@ contract ERC721A is IERC721A {
...
+
+            // If the next slot may not have been initialized (i.e. `nextInitialized == false`) .
+            if (prevOwnershipPacked & BITMASK_NEXT_INITIALIZED == 0) {
+                uint256 nextTokenId = tokenId + 1;
+                // If the next slot's address is zero and not burned (i.e. packed value is zero).
+                if (_packedOwnerships[nextTokenId] == 0) {
+                    // If the next slot is within bounds.
+                    if (nextTokenId != _currentIndex) {
+                        // Initialize the next slot to maintain correctness for `ownerOf(tokenId + 1)`.
+                        _packedOwnerships[nextTokenId] = prevOwnershipPacked;
```
Vectorized, line 583: Moving this into a function will cause overhead.

contracts/ERC721A.sol
cygaar: This is pretty clever, had to work it out on paper to see how this worked

test/ERC721A.test.js
```diff
@@ -225,6 +235,13 @@ const createTestSuite = ({ contract, constructorArgs }) =>
             it('adjusts owners balances', async function () {
               expect(await this.erc721a.balanceOf(this.from)).to.be.equal(1);
             });
+
+            it('startTimestamp updated correctly', async function () {
+              expect(this.timestampBefore).to.be.lt(this.timestampToMine);
+              expect(this.timestampAfter).to.be.gte(this.timestampToMine);
+              expect(this.timestampAfter).to.be.gt(this.timestampBefore);
```
cygaar, line 242: Isn't this statement redundant given the first two checks?

test/GasUsage.test.js
```diff
@@ -39,4 +39,15 @@ describe('ERC721A Gas Usage', function () {
       }
     });
   });
+
+  context('transferFrom', function () {
+    it('transfer to and fro', async function () {
```
cygaar, line 44: Typo: "from"

cygaar (approved):

cygaar:
Let's do one more poll to see how people feel about this

test/GasUsage.test.js
Vectorized: Ok, changed to "transfer to and from two addresses"

test/ERC721A.test.js
Vectorized: Oh yes, removed.

PR #275, opened 2022-05-10
Remove OpenZeppelin

Vectorized:
This PR will also make updating to diamond upgradeables easier.


contracts/ERC721A.sol
```diff
@@ -88,11 +92,11 @@ contract ERC721A is Context, ERC165, IERC721A {
     /**
      * @dev See {IERC165-supportsInterface}.
      */
-    function supportsInterface(bytes4 interfaceId) public view virtual override(ERC165, IERC165) returns (bool) {
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
```
cygaar, line 95: I would leave a comment in the code stating what these values are and where they're coming from

contracts/ERC721A.sol
```diff
@@ -276,7 +280,7 @@ contract ERC721A is Context, ERC165, IERC721A {
         bytes memory _data
     ) public virtual override {
         _transfer(from, to, tokenId);
-        if (to.isContract()) if(!_checkContractOnERC721Received(from, to, tokenId, _data)) {
+        if (to.code.length != 0) if(!_checkContractOnERC721Received(from, to, tokenId, _data)) {
```
cygaar, line 283: Is checking code.length equivalent to using the `EXTCODESIZE` opcode?

contracts/ERC721A.sol
```diff
@@ -5,11 +5,18 @@
...
+ * @dev ERC721 token receiver interface.
+ */
+interface __ERC721A__IERC721Receiver {
+    function onERC721Received(
+        address operator,
+        address from,
+        uint256 tokenId,
+        bytes calldata data
+    ) external returns (bytes4);
+}
```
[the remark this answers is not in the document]
cygaar: Any reason you're doing __ERC721A instead of just ERC721A?

contracts/ERC721A.sol
```diff
@@ -596,4 +601,44 @@ contract ERC721A is Context, ERC165, IERC721A {
...
+     * @dev Returns the message data (defaults to `msg.data`).
+     */
+    function _msgData() internal view virtual returns (bytes calldata) {
+        return msg.data;
+    }
+
+    /**
+     * @dev Converts a `uint256` to its ASCII `string` decimal representation.
+     */
+    function _toString(uint256 value) internal pure returns (string memory) {
```
cygaar, line 622: Might be worth throwing a test or two for this into our test suite

cygaar (commented):
Small comments otherwise lgtm

contracts/ERC721A.sol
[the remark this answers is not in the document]
Vectorized: Ok, let’s go with `ERC721A__`.

contracts/ERC721A.sol
Vectorized: Ok, i’ll leave a comment with the EIP165 link.

contracts/ERC721A.sol
Vectorized: yes

contracts/ERC721A.sol
Vectorized: Already added.

PR #280, opened 2022-05-14
Rename _msgSender

Vectorized:
`_msgSender()` is used for writing GSN compatible contracts.

When users import OZ and inherit from their contracts (e.g. Ownable), our `_msgSender()` will conflict with their definition.

I suggest to keep an overrideable function, with a different name, such that it won't conflict.


Vectorized:
On the same note, I'm also thinking of renaming `_toString` to something else, since it is a very common name that may conflict, and not really related to the core logic of ERC721A.

Vectorized:
I merge this first incase people are using forge to install (which doesn't pull from npm, but rather git).

PR #286, opened 2022-05-19
Added Forge Test 

Atarpara:
This PR is the smooth and fast the testing. This is my first PR to OOS Project.

I added forge test-cases into this project. You can find more info [Foundry-github](https://github.com/foundry-rs/).

Forge Installation : https://github.com/foundry-rs/foundry#installation

Run testcase: 

`forge install`
`forge test`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ node_modules/
 coverage
 coverage.json
 typechain
+out/
 
 #Hardhat files
 cache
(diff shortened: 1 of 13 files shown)


Vectorized:
We have considered adding Forge tests internally. 

In the end, we have decided to stick to only Hardhat tests for the following reasons:

- Maintain compatibility with the Upgradable version that is automatically transpiled by OpenZeppelin's transpiler (which is based on Hardhat).

- Enable easy code reuse to test difficult parts like `_startTokenId()` and the `ERC721AQueryable` functions. 

- We will only want to maintain one copy of tests. Imagine having to maintain two pieces of test logic for every change. 

We are fans of Foundry and appreciate your efforts. 

Unfortunately, we must be very selective on the PRs we accept for maintainability and compatibility.

Issue #289, opened 2022-05-22
Removal of ERC721AOwnersExplicit


Vectorized:
With the new `_initializeOwnershipAt(uint256 index)` function, users can directly initialize the ownership at any index. 

It is now manageable to write a public function to initialize the ownerships anywhere.

As such, we are planning to remove the `ERC721AOwnersExplicit` extension.

holic:
If I want to normalize/initialize ownership data after mint out (for better transfer gas), what's the recommended approach? Is there a better way than to iterate through tokens in batches and call `_initializeOwnershipAt`?

Vectorized:
@holic I would recommend restricting the quantity in the batch mint function to be below a reasonable amount (e.g. 30).

`_initializeOwnershipAt` can be useful if you need to mint a very large batch for the team (e.g. 200). 

You can now either mint anytime for the team (usually during the start or the end), and call `_initializeOwnershipAt` at say every multiple of 8, within the range minted for your team. Then, subsequent transfers will have more reasonable gas costs.

The recent improvements have the largest impact on second transfers (from about 29k gas to 24k gas).

The first transfer will see savings of around (2k gas if there is a need to initialize the next slot).

Note that the bulk of the first-time transfer costs for tokens minted in large batches is in the scanning of the ownership slots.

mulder1628:
Hi @Vectorized 

Can you provide example on how to go about calling `_initializeOwnershipAt` post mint (initial mint using ERC721A w/o this function) to reduce the scanning cost?  We have large gaps in the range of couple of hundreds.

Thank you for your help in advance!

PR #290, opened 2022-05-24
Remove ERC721ASetOwnersExplicit extension

cygaar:
The impetus for removing this extension is to simplify our codebase and remove code adds unnecessary complexity. This code was added as a fail safe when ERC721A was first written, but we've since added optimizations and other helpers that make this unnecessary. Like @Vectorized mentioned in the issue below, contract owners can use `_initializeOwnershipAt` at certain intervals to make transfers of large batch mints cheaper. Other than that scenario, it's not very necessary to explicitly set an ownership slot because the added cost of transferring an uninitialized slot isn't very high.

https://github.com/chiru-labs/ERC721A/issues/289


test/ERC721A.test.js
```diff
@@ -454,6 +454,21 @@ const createTestSuite = ({ contract, constructorArgs }) =>
...
+            const lastTokenId = this.addr3.expected.tokens[2];
+            expect(await this.erc721a.ownerOf(lastTokenId)).to.be.equal(this.addr3.address);
+            const tx1 = await this.erc721a.initializeOwnershipAt(lastTokenId);
+            expect(await this.erc721a.ownerOf(lastTokenId)).to.be.equal(this.addr3.address);
+            const tx2 = await this.erc721a.initializeOwnershipAt(lastTokenId);
+
+            // We assume the initialization worked due to less gas used by the 2nd txn
+            const receipt1 = await tx1.wait();
+            const receipt2 = await tx2.wait();
+            expect(receipt1.gasUsed.toNumber()).to.be.greaterThan(receipt2.gasUsed.toNumber());
```
cygaar, line 469: This is a best-efforst heuristic to determine whether or not the initialization call worked because we can't directly access `_packedOwnerships`

cygaar, line 469: This is a best-efforst heuristic to determine whether or not the initialization call worked because we can't directly access `_packedOwnerships`

Vectorized: We can access `_ownershipAt`, which is internal. 

But big brain trick haha.

Vectorized:
So many lines gone. Finally the itch got scratched. :)

LGTM.

PR #311, opened 2022-06-04
Add ERC-2309 support for mints during contract creation

Vectorized:
This PR will be refactored into providing a `_mintERC2309(to, quantity)` function, that is intended for efficient minting **during contract creation**.

The NatSpec will clearly document that **calling the function outside of contract creation, WILL make the contract non-compliant to the ERC721 standard** (even though it is still supported by some marketplaces on a de facto basis).

Todo:
- [x] Refactor into `_mintERC2309(to, quantity)`.


cygaar (approved):

cygaar:
Like we talked about, we should let users decide what they want to do for the limit, but add some comments/docs regarding the potential issues with higher limits

fulldecent:
I recommend that this PR be closed without further review.

ERC721a is advertised as "[an improved implementation of the [ERC-721] standard](https://www.erc721a.org)". And the [goal of the project](https://github.com/chiru-labs/ERC721A) is "to provide a fully compliant implementation of [ERC-721]".

This is the first sentence in both places.

This PR violates both of those guiding principles by removing ERC-721 compatibility. Or if you want to pedant, it violates the principle by allowing the customer to remove ERC-721 compatibility by quickly doing a thing.

---

If it is not clear to anybody, ERC-2309 is NOT compatible with ERC-721. The ERC-2309 specification is poorly named ("ERC-721 Consecutive Transfer Extension") whereas it is not an extension of ERC-721 but instead it is an alternative.

Vectorized:
@fulldecent Thanks. Closed.

Vectorized:
Imo, we can still provide a fully compliant implementation of ERC721 (by default),  
but still allow give users the flexibility to break standard if they wish to. 

Nothing is stopping users from copy-pasting the code and adding ERC2309 themselves.

Additionally, all marketplaces that support ERC2309 classify the contracts as ERC721.  
Are these marketplaces breaking standard? 

Previous versions of ERC721A didn't stop users from breaking logic if they directly write to the internal variables.

nftboi:
My biggest concern was marketplace support, but if looksrare and OS support w/ limits I'm all for this. 2309 has existed since 2019, but hasn't seen widespread adoption. I'd assume that's because there was no public, solid implementation. This is both public, vetted fully 721 compliant but can add 2309 and shine light on a standard which obviously adds tremendous efficiencies to initial mints. Imo everyone should use this for initial mints and adding it directly 721a will increase the likelihood that people do that. Creating a separate repo, or project just to add this additional event increases the likelihood the people DON'T use it. Regardless of the name, or if it's compliant or not compliant with the original 721 standard the amount of money that can be saved if every new projects uses this is astronomical. It's good for the NFT community and good for Ethereum. We should be doing everything we can to push this forward.

Vectorized:
I just thought of an idea to remain fully ERC721 compliant regardless to satisfy the purists/pendantics. ;) 

Instead of overriding, we simply include a `_mintERC2309(to, quantity)` function.

If called in the constructor, it is perfectly valid ERC721. The ERC721 standard says we don’t need to emitt `Transfer` events for NFTs created upon contract creation. It also doesn’t state that we cannot emit other kinds of events in the constructor, and doesn’t restrict what we can call the internal minting function. 

But for performance issues, we will omit the check for `this.code.length == 0` in the function.

JohnChangUK:
This is a really neat optimization! But as @Vectorized stated keeping compliant with the ERC721 standard holds more importance in my opinion.

fulldecent:
**Below are notes about allowing ER-2309 support broadly, even if behind an option.**

---

If this PR is accepted then the tagline of this project should change from:

> ERC721A is an improved implementation of the IERC721 standard that supports minting multiple tokens for close to the cost of one.

to

> ERC721A is an improved implementation of the IERC721 standard that supports minting multiple tokens for close to the cost of one, unless you override this one function in which case this is no longer an implementation of ERC721 and it is instead an implementation of ERC2903 which claims to be related to ERC721 but isn't.

That is a completely fair disclaimer of what is happening here.

---

@Vectorized The LLVM project, the best compiler, introduces itself as:

> The LLVM Project is a collection of modular and reusable compiler and toolchain technologies. Despite its name, LLVM has little to do with traditional virtual machines. The name "LLVM" itself is not an acronym; it is the full name of the project.

Implementation of IEEE-754 is not a core tenet of that project. So the LLVM argument is distinguishable from the ERC721a situation.

But instead, if there were a C compiler named IEEE754improved, and the tagline and project scope for that project was based on IEEE-754 support, then yes it would be inappropriate for such a project to have a feature that disabled IEEE-754 support.

---

> Breaking standards for practical reasons has been a common practice in the history of software engineering, and the benefits can outweigh the costs if done in a conscious, careful manner.

I agree that breaking standards is important. And I'm am the first person to tell people to NOT use ERC-721 or blockchain (instead, most companies should use traditional software with more transparency and assurance features).

But the discussion here is not on the principle of which software in the universe should exist. Instead, we are narrowly focused on which software should be part of the ERC721a project.

We have a guiding principle for this. And this is the project scope. The project scope is clearly defined, it is highly visible and it is inside the name of this project. And that scope does exclude ERC-2309. Therefore this PR should be closed without merging or further consideration.

---

> This is both public, vetted fully 721 compliant but can add 2309 and shine light on a standard which obviously adds tremendous efficiencies to initial mints.

The goal of this project/repository is not to promote new competing standards against ERC-721.[Open source does not mean your company can depend on the world’s best labor without spending money.](https://twitter.com/fulldecent/status/1529484114071588865).

Quite the contrary, this repository is premised on continuing support of ERC-721, the standard for non-fungible tokens.

---

> Creating a separate repo, or project just to add this additional event increases the likelihood the people DON'T use it. Regardless of the name, or if it's compliant or not compliant with the original 721 standard the amount of money that can be saved if every new projects uses this is astronomical. It's good for the NFT community and good for Ethereum. We should be doing everything we can to push this forward.

For a detailed discussion on why fragmentation of non-fungible token standards is terrible for the community, please refer to the discussions in EIP issues 721, 841 and proceedings from other meetings on finalization. I have cataloged those for you at at https://phor.net.

Quite plainly, if there had been multiple competing standards 4 years ago (ERC-721's finalization birthday is in two weeks) then very people would know about NFTs today. Fragmenting now is much worse than 4 years ago.

There are many other practical ways to reduce dollar cost of deploying contracts which this margin is too narrow to contain.

Vectorized:
We will clearly document in the NATSPEC that usage of the `_mintERC2309` function outside of the constructor WILL break the ERC721 standard.

fulldecent:
I reserve further comments until I can see the full proposed PR.

Vectorized:
@fulldecent

Vectorized:
@fulldecent

Vectorized:
Last call for comments. Merging in a day or few.

contracts/ERC721A.sol
```diff
@@ -464,6 +466,10 @@ contract ERC721A is IERC721A {
      * - `to` cannot be the zero address.
      * - `quantity` must be greater than 0.
      *
+     * Note: some marketplaces may have difficulties fully registering a very large `quantity`.
```
fulldecent, line 469: This note is not applicable for this function.

OpenSea and LooksRare support this `_mint` implementation with any quantity of tokens so long as the transaction is posted to the ledger. This is because this transaction would fail with out of gas if trying too large a quantity.

contracts/ERC721A.sol
```diff
@@ -504,6 +510,65 @@ contract ERC721A is IERC721A {
         _afterTokenTransfers(address(0), to, startTokenId, quantity);
     }
 
+    /**
+     * @dev Mints `quantity` tokens and transfers them to `to`.
+     *
+     * This function is intended for efficient minting during contract creation.
```
fulldecent, line 516:
```suggestion
     * This function is intended for efficient minting only during contract creation.
```

contracts/ERC721A.sol
```diff
@@ -504,6 +510,65 @@ contract ERC721A is IERC721A {
...
+     *
+     * It emits only one {ConsecutiveTransfer} as defined in
+     * [ERC2309](https://eips.ethereum.org/EIPS/eip-2309),
+     * instead of a sequence of {Transfer} event(s).
+     *
+     * Calling this function outside of contract creation WILL break the ERC721 standard.
```
fulldecent, line 522:
```suggestion
     * Calling this function outside of contract creation WILL make your contract non-compliant with the ERC721 standard.
```

contracts/ERC721A.sol
```diff
@@ -504,6 +510,65 @@ contract ERC721A is IERC721A {
...
+     * Emits a {ConsecutiveTransfer} event.
+     */
+    function _mintERC2309(address to, uint256 quantity) internal {
+        uint256 startTokenId = _currentIndex;
+        if (_addressToUint256(to) == 0) revert MintToZeroAddress();
+        if (quantity == 0) revert MintZeroQuantity();
+
+        _beforeTokenTransfers(address(0), to, startTokenId, quantity);
+
+        // Overflows are incredibly unrealistic.
```
fulldecent, line 544: ⚠️  Security issue.

Overflows are incredibly easy and they result in data corruption across the bitfield, and inconsistent results in `_afterTokenTransfers` versus the quantity actually minted.

For example, this function allows to effectively mint negative one token by passing a value of 2^255.

contracts/mocks/ERC721AWithERC2309Mock.sol
```diff
@@ -0,0 +1,29 @@
...
+// ERC721A Contracts v4.0.0
+// Creators: Chiru Labs
+
+pragma solidity ^0.8.4;
+
+import '../ERC721A.sol';
+
+/**
+ * @dev Mock for testing and benchmarking purposes.
+ * Calling `_mintERC2309` outside of contract creation breaks the ERC721 standard.
```
fulldecent, line 11:
```suggestion
 * Calling `_mintERC2309` outside of contract creation makes your contract non-compliant with the ERC721 standard.
```

test/GasUsage.test.js
```diff
@@ -40,6 +40,20 @@ describe('ERC721A Gas Usage', function () {
     });
   });
 
+  it('mintOneERC2309', async function () {
+    let contract = await deployContract('ERC721AWithERC2309Mock', ['Azuki', 'AZUKI']);
+    await contract.mintOneERC2309(this.owner.address);
```
fulldecent, line 45: This test case is not following the ERC-721 standard.

Please update all test cases to only demonstrate uses cases with are within the project scope, i.e. fully supporting the ERC-721 standard.

test/GasUsage.test.js
```diff
@@ -40,6 +40,20 @@ describe('ERC721A Gas Usage', function () {
...
+    await contract.mintOneERC2309(this.owner.address);
+    await contract.mintOneERC2309(this.addr1.address); 
+  });
+
+  it('mintTenERC2309', async function () {
+    let contract = await deployContract('ERC721AWithERC2309Mock', ['Azuki', 'AZUKI']);
+    await contract.mintTenERC2309(this.owner.address);
```
fulldecent, line 52: This test case is not following the ERC-721 standard.

Please update all test cases to only demonstrate uses cases with are within the project scope, i.e. fully supporting the ERC-721 standard.

fulldecent (changes requested):

fulldecent:
This is my complete review for the pull request. ^^^

contracts/ERC721A.sol
Vectorized: Added a max `quantity` limit of 5000 per call to `_mintERC2309`.

contracts/ERC721A.sol
Vectorized: Manual emprical testing on their Rinkeby testnet marketplaces shows that they are incappable of supporting any quantity of tokens though.

test/GasUsage.test.js
Vectorized: Refactored to have all calls to `_erc2309` in the constructor for the tests.

Vectorized:
@fulldecent Thank you. The request changes have been added.

contracts/ERC721A.sol
fulldecent: Please post the testing tx you are talking about

fulldecent:
The test cases are still showing out-of-scope non-ERC-721 use cases.

If these test cases are desired, they should be explained that if you do that you are non-compliant with ERC-721.

Vectorized:
@fulldecent The non-compliant code is included only for gas comparisons. I have added comments to clarify that.

Issue #83, opened 2022-02-09
Add transferFrom gas estimates for various situations


fulldecent:
This is an important consideration of whether to use this implementation.

fulldecent:
Here's a start 

https://twitter.com/fulldecent/status/1491506122565595141

Did this manually, we can make good tooling for this.

Vectorized:
The more popular the project, the more gas ERC721A can save. 

The gas savings during mint (high BASEFEE period) can be way more than  
the extra overhead incurred when transferring tokens (low BASEFEE period).

A general rule-of-thumb to minimize gas is to transfer in ascending tokenId order.

Also, it has the indirect effect of discouraging dumping (promoting hodling), especially for those who mint large batches. ;)

0xsudo-eth:
@fulldecent  @Vectorized  check [this issue](https://github.com/chiru-labs/ERC721A/issues/104), if we good with the proposal I can do

alephao:
Hello, I made a benchmark that is more precise than the one showing here and compares gas usage with other implementations. It compares not only `mint` but also `transferFrom` on many cases and all other methods so the developers have a better picture of what the gas usage will be when using this or any other contract.

Feel free to grab the results or link the codebase, I think it's more trustworthy than the one showing on README since all comparisons are made using the same process, and by someone not involved in this project.

https://github.com/alephao/solidity-benchmarks
https://github.com/alephao/solidity-benchmarks/blob/main/ERC721.md

caffeinum:
I wonder how it scales in the future? Most of the benchmarks are only interested in 100 tokens; but when Ethereum NFT is the base for (m)etaverse, and we need collections of billions NFTs, some people owning millions of them, how many gas transfer will eat? Even if the block gas limit is raised to 100_000_000 gas, is it possible to being "locked out" of your tokens?

Vectorized:
@caffeinum If you mint at the block gas limit (about 16K tokens for now on ETH), you won’t be able to directly transfer the last token in the batch. This is because the LOG4 opcode (1875 gas) emitted per mint costs less than a SLOAD opcode (2100 gas). 

You can however, transfer a token in the middle of the batch to initialize it so that the token at the end can be transferred within the block gas limit. 

In an upcoming PR #272, we will have a function to directly initialize a token without transferring it.

basememara:
[...]

PR #314, opened 2022-06-05
Add ERC721ASplitMint extension

caffeinum:
Automatically split mints into batches to save gas on subsequent transfers

Simplest usage example:
```solidity
import "../ERC721A.sol";
import "../extensions/ERC721ASplitMint.sol";

contract ERC721ASplitMintMock is ERC721A, ERC721ASplitMint {
    constructor(string memory name_, string memory symbol_) ERC721A(name_, symbol_) {}

    function _mintBatchSize() override internal pure returns (uint256) {
        // we require user to override this function so that he explicitly knows the batch size
        return DEFAULT_MINT_BATCH_SIZE;
    }

    function mint(uint256 quantity) public {
       // note we're using _splitMint instead of _safeMint
        _splitMint(msg.sender, quantity);
    }

}

```

### DIFF
--- a/contracts/extensions/ERC721ASplitMint.sol
+++ b/contracts/extensions/ERC721ASplitMint.sol
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: MIT
+// ERC721A Contracts v4.0.0
+// Creator: @caffeinum
+
+pragma solidity ^0.8.4;
+
+import '../ERC721A.sol';
+
+/**
+ * @title ERC721A Split Mint
+ * @dev Splits mint into batches by _mintBatchSize tokens. 
+ */
+abstract contract ERC721ASplitMint is ERC721A {
+
+    uint256 constant DEFAULT_MINT_BATCH_SIZE = 10;
+
+    /**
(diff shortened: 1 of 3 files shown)


caffeinum:
The idea follows this comment
https://github.com/chiru-labs/ERC721A/issues/83#issuecomment-1125686598

Batched mints are the way to configure default behavior for ERC721A so that it's future-proof against people who gonna launch collections of size of billions of tokens.

I would even say splitMint should be the default one, and safeMint should be renamed to unboundedSafeMint, but that's up for discussion.

Another two decisions where I would like feedback is:
- If you import ERC721ASplitMint extension, should splitMint override safeMint automatically? I would say yes, but in this implementation I left it optional just in case
- Should we provide default value for _mintBatchSize or user should be required to type it in?

Vectorized:
Right now, it seems like OpenSea's Rinkeby version cannot handle 10k NFTs being minted in a single transaction.

Let me know if this extension can help alleviate the NFTs getting stucked.

Imo, most if not all collections currently impose a reasonable limit of the quantity of NFTs that can be publicly minted in a single tx.

Issue #322, opened 2022-06-07
Allowing general storage hitchhiking


cxkoda:
Currently, ERC721A stores `startTimestamp` efficiently with each `TokenOwnership` in storage. While this is convenient for tokenomics, for some projects it would be nice to use the respective 64 bits for something else. More flexibility could be achieved by splitting the data providing logic (currently `block.timestamp`) into an own internal virtual function that is called on each transfer and renaming `TokenOwnership.startTimestamp` to something more generic like `TokenOwnership.extraData`.

Before submitting a PR, I wanted to check if this is something that you would like to support or if this is beyond the scope of ERC721A.

cxkoda:
Oh sorry, I should have been more explicit. I was only talking about lazy-initializable data, e.g. using the blocknumber, or parts of the blockhash instead of the timestamp.

So on transfers having something like
```js
...
_packedOwnerships[startTokenId] =
    _addressToUint256(to) |
    (_extraData(from, to, previousExtraData) << BITPOS_START_TIMESTAMP) |
    (_boolToUint256(quantity == 1) << BITPOS_NEXT_INITIALIZED);
...

function _extraData(address from, address to, uint64 previousExtraData) internal view virtual returns (uint256) {
   return block.number;
}
```

My usecase is for generative art, where we want to generate and store a random seed on mint. Using this scheme, we would generate a proto-seed for each batch, and calculate the actual seed of each token later via something like `keccak(proto, tokenId)`. I'd override the above function to generate something new on mint, and reuse the previous value on transfers.

cxkoda:
Yes this will absolutely be a breaking change because `startTimestamp` needs to be renamed. Maybe to something like `extraData`? It would be very similar to the `aux` for `addressData` in a sense (only lazily initialized). 

I submitted a quick draft implementation in #323 to show you more in more detail how I intended this. Also talking code is easier most times. IMO, it can be added pretty straightforwardly without any overhead in terms of gas (the report remained completely unchanged on my end) and only small overhead in terms of maintainability (2 extra functions).

But I understand that this might be a feature that you don't necessarily want to support/maintain upstream. In that case I'll gladly fork.

cxkoda:
Awesome! Then I'll clean it up a bit and add some more comments and tests in the next days.

Regarding release, I totally understand! Happy if it gets considered for the next major version.

Vectorized:
I'm thinking if 24 bits of `extraData` is enough. That's 16777215 values.

[...]

PR #323, opened 2022-06-08
Add a generic interface for ownership storage hitchhiking

cxkoda:
A quick draft implementation for #322 "Allowing general storage hitchhiking".


contracts/ERC721A.sol
```diff
@@ -633,6 +648,7 @@ contract ERC721A is IERC721A {
             _packedOwnerships[tokenId] =
                 _addressToUint256(from) |
                 (block.timestamp << BITPOS_START_TIMESTAMP) |
+                (_extraData(from, address(0), _unpackExtraData(prevOwnershipPacked)) << BITPOS_EXTRA_DATA) |
                 BITMASK_BURNED |
                 BITMASK_NEXT_INITIALIZED;
```
cxkoda, line 653: One of my new tests for the behavior of  `_extraData` failed for burns on my machine. I was able to track it back to the `_addressToUint256(from)` part in the above code. Replacing it with `(_addressToUint256(from) & ((1 << 160) - 1)) |`, so zeroing the leading 96bits solved the issue. I have no idea why `_addressToUint256` seems to produce dirty `uint256`s though. @Vectorized

Vectorized: Ok, i figured out.

This is because `extraData` is not at the last 24 bits of the uint256.

For efficiency reasons, and for backwards storage compatibility reasons (we have a upgradeable repo), 
we will need to use the following layout:
```solidity
    // Bits Layout:
    // - [0..159]   `addr`
    // - [160..223] `startTimestamp`
    // - [224]      `burned`
    // - [225]      `nextInitialized`
    // - [232..255] `extraData`
```

Vectorized: Hmm... strange.

Anyway, i also encountered the same issue. I've edited some stuff to solve it.

Vectorized:
Ok, did some changes:

1. Merged the latest changes
2. Mask the address in _packOwnershipData` to prevent dirty bits.
3. Changed `extraData` to use the last 24 bits.
4. Changed all zero address checks to the regular version (for safety reasons).

contracts/ERC721A.sol
```diff
@@ -270,6 +275,8 @@ contract ERC721A is IERC721A {
      */
     function _packOwnershipData(address owner, uint256 flags) private view returns (uint256 value) {
         assembly {
+            // Mask `owner` to the lower 160 bits, in case the upper bits somehow aren't clean.
+            owner := and(owner, 0xffffffffffffffffffffffffffffffffffffffff)
```
cygaar, line 279: I kinda like having `(1 << 160) - 1` as a top level constant instead of 40 f's. I know you like hex values for assembly though, so you can have the final call

contracts/ERC721A.sol
Vectorized: It looks like the compiler didn't clean the bits.

I have added some changes to ensure that.

contracts/ERC721A.sol
```diff
@@ -693,6 +697,41 @@ contract ERC721A is IERC721A {
...
+    function _nextExtraData(
+        address from,
+        address to,
+        uint256 prevOwnershipPacked
+    ) internal view virtual returns (uint256) {
+        uint24 previousExtraData;
+        assembly {
+            previousExtraData := shr(BITPOS_EXTRA_DATA, prevOwnershipPacked)
+        }
+        return uint256(_extraData(from, to, previousExtraData)) << BITPOS_EXTRA_DATA;
```
cygaar, line 712: Is it more expensive to return a uint24 here? In terms of readability, it's more clear what the value here should be if we return uint24.

Vectorized: This private function is for returning the shifted extra data.

PR #334, opened 2022-06-12
Small gas optimization for burn/transfers/mint 

kadenzipfel:
first commit:
- saves 9 gas per burn
- saves 3 gas per transfer
- increases deployment cost but imo worth it

second commit:
- saves 6*(quantity-1)-8 gas per mint
- imo super worth the additional upfront cost

https://www.diffchecker.com/rYLkEpYd

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -494,10 +494,11 @@ contract ERC721A is IERC721A {
                 (block.timestamp << BITPOS_START_TIMESTAMP) |
                 (_boolToUint256(quantity == 1) << BITPOS_NEXT_INITIALIZED);
 
-            uint256 offset;
+            uint256 offset = startTokenId;
+            uint256 end = quantity + startTokenId;
             do {
-                emit Transfer(address(0), to, startTokenId + offset++);
-            } while (offset < quantity);
+                emit Transfer(address(0), to, offset++);
+            } while (offset < end);
 
             _currentIndex = startTokenId + quantity;
         }
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -266,6 +266,16 @@ contract ERC721A is IERC721A {
     }
 
     /**
+     * @dev Packs ownership data into a single uint256.
+     */
+    function _packOwnershipData(address owner, uint256 flags) private view returns (uint256 value) {
+        assembly {
+            // `owner | (block.timestamp << BITPOS_START_TIMESTAMP) | flags`.
+            value := or(owner, or(shl(BITPOS_START_TIMESTAMP, timestamp()), flags))
+        }
+    }
+
+    /**
      * @dev See {IERC721-ownerOf}.
      */
     function ownerOf(uint256 tokenId) public view override returns (address) {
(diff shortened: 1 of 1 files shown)


Vectorized:
I think the burn part shouldn't be in assembly. Gas savings not really worth the extra function.

Maybe we can consider using assembly for events.

```solidity
uint256 end = startTokenId + quantity;

// Use assembly to save around 30+ gas per iteration.
assembly {
    let index := startTokenId
    // emit Transfer(from, to, index)
    // The magic number is `keccak256(bytes("Transfer(address,address,uint256)"))`.
    log4(0, 0, 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef, 0, to, index)
    index := add(index, 1)
    for {} iszero(eq(index, end)) { index := add(index, 1) } {
        log4(0, 0, 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef, 0, to, index)
    }
}

_currentIndex = end;
```

kadenzipfel:
> I think the burn and transfer parts shouldn't be in assembly. Gas savings not really worth the extra function.

if we're thinking in terms of overall costs, offsetting the deployment cost would only require ~500 transfers, ~166 burns or a combination of the two. for most collections, 500 transfers is trivial and thus imo super worth it

if we're thinking in terms of devex, imo it's much more readable to have the extra function describing the process

> Plus, it may make it harder to extend (e.g. for the storage hitchhiking PR https://github.com/chiru-labs/ERC721A/pull/323 ).

it's quite easy to extend actually and in the case of the storage hitchhiking PR the gas savings are even more significant (41 gas for burn, 6 gas for transfer)
see: https://github.com/kadenzipfel/ERC721A/blob/d69b41f83614db42ec9dec97ce3a8d68cebfb843/contracts/ERC721A.sol#L569
https://www.diffchecker.com/xUhWcd6r

> Maybe we can consider using assembly for emitting the Transfer event in _mint.
+1

Vectorized:
@kadenzipfel Not really a fan of functions for code that is used only in one spot.

contracts/ERC721A.sol
```diff
@@ -575,6 +573,15 @@ contract ERC721A is IERC721A {
...
 
+    /**
+     * @dev Packs ownership data into single uint
+     */
+    function _packTransferOwnershipData(address to) private view returns (uint256 packedData) {
+        assembly {
+            packedData := or(to, or(shl(BITPOS_START_TIMESTAMP, timestamp()), BITMASK_NEXT_INITIALIZED))
+        }
+    }
+
```
Vectorized, line 584: Refactor it into:

```solidity
/**
 * @dev Packs ownership data.
 */
function _packOwnershipData(address owner, uint256 flags) private view returns (uint256 value) {
    assembly {
        value := or(owner, or(shl(BITPOS_START_TIMESTAMP, timestamp()), flags))
    }
}
```

And use it across `_mint`, `_transfer`, `_burn`.

Should give similar results.

Vectorized: Somehow the compiler is so stubborn for this case that it needs some assembly to whip it into place.

cygaar:
I'm renaming the PR to: "Add more gas optimizations"

kadenzipfel:
updated gas savings: https://www.diffchecker.com/BS9LslEP

Vectorized:
LGTM.

Vectorized (approved):
LGTM

PR #337, opened 2022-06-14
Some minor optimizations and refactors

Vectorized:
This will make using assembly more worth it imo.  
Gas savings are in the tens.

@cygaar

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -643,7 +643,7 @@ contract ERC721A is IERC721A {
         // Clear approvals from the previous owner.
         assembly {
             if approvedAddress {
-                // This is equivalent to delete `_tokenApprovals[tokenId]`.
+                // This is equivalent to `delete _tokenApprovals[tokenId]`.
                 sstore(approvedAddressSlot, 0)    
             }
         }
@@ -718,7 +718,7 @@ contract ERC721A is IERC721A {
         // Clear approvals from the previous owner.
         assembly {
             if approvedAddress {
-                // This is equivalent to delete `_tokenApprovals[tokenId]`.
+                // This is equivalent to `delete _tokenApprovals[tokenId]`.
                 sstore(approvedAddressSlot, 0)    
             }
         }
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -580,15 +580,34 @@ contract ERC721A is IERC721A {
     }
 
     /**
-     * @dev Zeroes out _tokenApprovals[tokenId]
+     * @dev Returns the storage slot and value for the approved address of `tokenId`.
      */
-    function _removeTokenApproval(uint256 tokenId) private {
-        mapping(uint256 => address) storage tokenApprovalPtr = _tokenApprovals;
+    function _getApprovedAddress(
+        uint256 tokenId
+    ) private view returns (uint256 approvedAddressSlot, address approvedAddress) {
+        mapping(uint256 => address) storage tokenApprovalsPtr = _tokenApprovals;
         assembly {
+            // Compute the slot and load it. This is equivalent to `approvedAddress[tokenId]`.
             mstore(0x00, tokenId)
-            mstore(0x20, tokenApprovalPtr.slot)
-            let hash := keccak256(0, 0x40)
-            sstore(hash, 0)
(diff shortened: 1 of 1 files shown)


contracts/ERC721A.sol
```diff
@@ -580,15 +580,36 @@ contract ERC721A is IERC721A {
...
+    /**
+     * @dev Returns whether the `approvedAddress` is equals to `from` or `msgSender`.
+     */
+    function _isOwnerOrApproved(
+        address approvedAddress,
+        address from, 
+        address msgSender
+    ) private pure returns (bool result) {
+        assembly {
+            // Mask `from` to the lower 160 bits, in case the upper bits somehow aren't clean.
```
cygaar, line 609: I think we should also mask `msgSender` here

contracts/ERC721A.sol
```diff
@@ -580,15 +580,36 @@ contract ERC721A is IERC721A {
...
+        }
+    }
+
+    /**
+     * @dev Returns whether the `approvedAddress` is equals to `from` or `msgSender`.
+     */
+    function _isOwnerOrApproved(
+        address approvedAddress,
+        address from, 
+        address msgSender
```
cygaar, line 606: Do we need to make this an input param, or can we just call `msgSenderERC721A()` directly? Seems like the goal of this is to always validate against msgSender, so there's no point in passing it in as an arg

contracts/ERC721A.sol
Vectorized: Masked.

cygaar:
Will take one more pass in the AM tomorrow

contracts/ERC721A.sol
Vectorized: Yes, `_msgSenderERC721A()` may be overridden for GSN contracts.

contracts/ERC721A.sol
```diff
@@ -580,87 +632,50 @@ contract ERC721A is IERC721A {
...
-                    if (nextTokenId != _currentIndex) {
-                        // Initialize the next slot to maintain correctness for `ownerOf(tokenId + 1)`.
-                        _packedOwnerships[nextTokenId] = prevOwnershipPacked;
-                    }
-                }
-            }
+    /**
+     * @dev Casts the boolean to uint256 without branching.
+     */
+    function _toUint256(bool value) private pure returns (uint256 result) {
```
cygaar, line 675: Why rename the function?

Vectorized: Shorter.

cygaar: I still think we should keep the original name. When reading the calling code, you have to double take to make sure you're converting a bool to int

contracts/ERC721A.sol
```diff
@@ -684,26 +699,26 @@ contract ERC721A is IERC721A {
...
-            bool isApprovedOrOwner = (_msgSenderERC721A() == from ||
-                isApprovedForAll(from, _msgSenderERC721A()) ||
-                approvedAddress == _msgSenderERC721A());
+        (uint256 approvedAddressSlot, address approvedAddress) = _getApprovedAddress(tokenId);
 
-            if (!isApprovedOrOwner) revert TransferCallerNotOwnerNorApproved();
-        }
+        if (approvalCheck)
+            if (!_isOwnerOrApproved(approvedAddress, from, _msgSenderERC721A()))
+                if (!isApprovedForAll(from, _msgSenderERC721A())) revert TransferCallerNotOwnerNorApproved();
```
cygaar, line 707: Can you combine this with the if-statement on 706? The formatting looks weird right now.

Vectorized: This is intentional for performance.

contracts/ERC721A.sol
Vectorized: Ahh… ok then.

contracts/ERC721A.sol
Vectorized: 26

cygaar: Can you add a comment here then? I know people are going to ask about this otherwise.

cygaar (approved):

contracts/ERC721A.sol
Vectorized: Ok, added

PR #347, opened 2022-06-20
Assembly mint loop

Vectorized:
Just letting it sit here. 

May or may not decide to merge. 

OpenSea Rinkeby is able to detect the events as per normal. 

Saves 36 gas for the first mint and 28 gas per additional iteration 
(`28 / 1958 ≈ 1.4%` savings per additional iteration).

The variables and statements have been carefully arranged to ensure that the compiler is as well behaved as possible. (actually, it can perform better than 100% handwritten Yul cuz Solidity provides high level semantics that can aid optimizations).

**Before:**
```
|  ERC721AGasReporterMock  ·  mintOne          ·      56201  ·      90401  ·      59621  ·
···························|···················|·············|·············|·············|
|  ERC721AGasReporterMock  ·  mintTen          ·      73828  ·     108028  ·      97768  ·
```

**After:**
```
|  ERC721AGasReporterMock  ·  mintOne          ·      56165  ·      90365  ·      59585  ·
···························|···················|·············|·············|·············|
|  ERC721AGasReporterMock  ·  mintTen          ·      73540  ·     107740  ·      97480  ·
```

Some trivia:

- `iszero(eq(tokenId, end))` is more efficient than `lt(tokenId, end)`   
because of how the compiler implicitly adds a `iszero` for the `lt`, and removes the `iszero` for the `eq`.

- Moving the zero address check after the loop helps avoid an implicit mask.

- Rewriting the `_packedAddressData[to] = ...` in assembly actually costs more gas.

- Moving `uint256 toMasked;` after `uint256 end = startTokenId + quantity;` actually costs more gas.


contracts/ERC721A.sol
```diff
@@ -508,11 +512,33 @@ contract ERC721A is IERC721A {
...
+                for {
+                    let tokenId := add(startTokenId, 1)
+                } iszero(eq(tokenId, end)) {
+                    tokenId := add(tokenId, 1)
+                } {
+                    // Emit the `Transfer` event. Similar to above.
+                    log4(0, 0, TRANSFER_EVENT_SIGNATURE, 0, toMasked, tokenId)
+                }
+            }
+            if (toMasked == 0) revert MintToZeroAddress();
```
pcaversaccio, line 541: question here: does we really need this check? Some background: One of the reasons that the zero checks were put there in the first place was that initially, Solidity didn't verify the length of `msg.data`, and thus it was possible for a function to be called with missing arguments. If an address argument was missing it would be read as the zero address, and these checks served the purpose of protecting from that kind of accident. Nowadays Solidity inserts checks that `msg.data` is big enough, so this kind of error is not really possible anymore. Second, what if the address is one of e.g. the pre-compiles (0x00...1 etc). Why is this accepted? Just wanted to raise this issue quickly.

contracts/ERC721A.sol
```diff
@@ -72,6 +72,11 @@ contract ERC721A is IERC721A {
     // is required to cause an overflow, which is unrealistic.
     uint256 private constant MAX_MINT_ERC2309_QUANTITY_LIMIT = 5000;
 
+    // The `Transfer` event signature is given by:
+    // `keccak256(bytes("Transfer(address,address,uint256)"))`.
+    uint256 private constant TRANSFER_EVENT_SIGNATURE =
```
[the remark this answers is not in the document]
pcaversaccio: also, to be a little bit picky here, `keccak256` returns `bytes32` and therefore you could change the type to `bytes32` instead of `uint256`.

contracts/ERC721A.sol
Vectorized: This is due to the EIP-721 spec.
https://eips.ethereum.org/EIPS/eip-721

> NFTs assigned to the zero address are considered invalid

So technically, according to the spec, an NFT can be owned by 0x000...1 and be considered valid.

pcaversaccio: I do understand why the authors have chosen this but still my argument holds true. Since it's a general library, I assume it's best to exactly follow the specs, but yeah, am not really happy about the zero address checks since they indicate some artificial security which does actually not exist. So if you keep this check, we could move it into assembly, no?

Vectorized: Moving it into assembly will mean assembloored reverts, which breaks aesthetics here (unlike Solmate's STL where everything is in assembly).

In this case, the gas numbers are the same for both assembloored reverts and plain ol' Solidity reverts.

pcaversaccio: I understand - my thoughts were not related to the gas but more to the consistency since the check belongs to the minting itself and if we aim for assembly-based minting it should be in the same Yul block.

cygaar (approved):

Issue #358, opened 2022-07-08
Some base functions are not virtual and thus can't be overridden by default. 


Sdocquir:
Some base functions are not virtual and thus can't be overridden by default. 
For example, ownerOf(), balanceOf(), etc.
Is there any reason behind that design decision ? Openzeppelin has theirs all virtual to allow the most amount of extensions/customizations.

I have a use case for an extension to ERC721A that would involve modifying the logic post-mint. I'm fine with simply changing those functions to virtual in my source code, but was also wondering if there's more to that design decision that I might be missing.

Vectorized:
Just a disclaimer, we will very most likely reject extension proposals though.

Sdocquir:
We're writing this extension for our project.
I understand modifying those base functions could be considered to be outside the purview of an extension and might warrant maintaining a separate standard altogether, but I figured it wouldn't hurt to start with a proposal and get community feedback.

No worries if/when it gets rejected.

PR #359, opened 2022-07-08
Make all public and internal functions virtual

Vectorized:
For #358 

To be honest, part of my OCD is triggered when functions that are dangerous to be overridden are marked virtual.

But a wise comment from a comrade soothed me. 
Screen Shot 2022-07-08 at 11 43 15 PM

https://twitter.com/rage_pit/status/1545423297667555328?s=20&t=o7Ijl8TrbjRSpXshoqIozA

OpenZeppelin and Solmate recently marked their functions as virtual too.

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -128,7 +128,7 @@ contract ERC721A is IERC721A {
     /**
      * @dev Returns the next token ID to be minted.
      */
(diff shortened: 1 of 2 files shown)


cygaar (approved):

sillytuna:
I came here specifically to ask for this.

Reasoning - quirky token id system needs extra functions overriding. I'd rather do it in a way which keeps me in sync.

PR #370, opened 2022-07-12
ERC4907

Vectorized:
https://eips.ethereum.org/EIPS/eip-4907

@emojidao @0xanders

Notes:

- File is called `ERC4907A.sol` because it inherits from `ERC721A.sol`. 

- The ERC-4907 standard is a **fully valid extension** of ERC-721.

- User info is **retained** after transfer, unlike [the reference implementation](https://github.com/ethereum/EIPs/tree/master/assets/eip-4907) which resets it after each transfer to a different address.
  
  - The standard permits the user info to remain the same upon transfer. 
  
  - Retaining the user info by default saves more gas overall. 
  
  - The new owner can choose to reset the user if they want.

  - Good for use cases such as retaining user info while sending the token to a personal cold wallet.
  
  - The `_beforeTokenTransfers` function can be overriden by devs to alter this behavior.

- Unlike the reference implementation, we do not have an internal mapping of structs.

  - Instead, we provide a `_explicitUserOf` which returns the address of the user, regardless of expiry status.

Some minor changes to `ERC721A.sol`:

- Add `_isApprovedOrOwner(address spender, uint256 tokenId)`. 

- Renamed private function `_isOwnerOrApproved` to `_isSinglyApprovedOrOwner` for more clarity.


contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * @dev Returns the user address for `tokenId`.
+     * The zero address indicates that there is no user or if the user is expired.
+     */
+    function userOf(uint256 tokenId) public view returns (address) {
+        uint256 packed = _packedUserInfo[tokenId];
+        assembly {
+            let expiry := shr(_BITPOS_EXPIRES, packed)
+            // `expires >= block.timestamp ? 1 : 0`.
+            let notExpired := iszero(lt(expiry, timestamp()))
```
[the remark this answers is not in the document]
Vectorized: Added comments.

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,135 @@
...
+                // `!(block.timestamp > expires) ? 1 : 0`.
+                iszero(
+                    gt(
+                        timestamp(),
+                        shr(_BITPOS_EXPIRES, packed) // `expires`.
+                    )
+                )
+            )
+        }
+        return address(uint160(packed));
```
obatirou, line 70: What do you think about dropping the negation ?
```solidity 
      function userOf(uint256 tokenId) public view returns (address) {
         uint256 packed = _packedUserInfo[tokenId];
         assembly {
             // Branchless `packed *= expires > block.timestamp  ? 1 : 0`.
             packed := mul(
                 packed,
                 // `expires > block.timestamp ? 1 : 0`.
                 gt(
                     shr(_BITPOS_EXPIRES, packed), // `expires`.
                     timestamp()
                 )
             )
         }
         return address(uint160(packed));
```

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,135 @@
...
+            // clear any non-zero excess upper bits.
+            mayNeedClearing := and(
+                eq(quantity, 1),
+                // `!(from == address(0) || from == to)`.
+                iszero(
+                    or(
+                        // Whether it is a mint (i.e. `from == address(0)`).
+                        iszero(and(from, _BITMASK_ADDRESS)),
+                        // Whether the owner is unchanged (i.e. `from == to`).
+                        eq(and(from, _BITMASK_ADDRESS), and(to, _BITMASK_ADDRESS))
```
0xPhaze, line 121: Is masking really required here?

Vectorized: I am paranoid. 

Ever since after encountering the bits not getting cleared previously.

Not sure if Solidity will clear the upper bits before passing the addresses in.

In the case where the bits are actually cleared... maybe the compiler will optimize it away (like previously observed).

contracts/extensions/ERC4907A.sol
Vectorized: Lemme think hard about it. 

The negation is to be faithful to the reference implementation.

Vectorized: Ok, i did something. If you know you know.

Vectorized: I did a little trick. ;)

Now it’s faithful to the reference.

contracts/extensions/ERC4907A.sol
0xPhaze: Hm, the function should only accept `address`, but I feel your paranoia. Had something similar happen.

contracts/extensions/ERC4907A.sol
Vectorized: **Explanation**

First, we know:

`packed = (expires << 160) | user`

The straightforward way of implementing `block.timestamp <= expires` in is:

`!(block.timestamp > (packed >> 160))` which in Yul, is: `iszero(gt(timestamp(), shr(160, packed)))`


Now, notice that if `user != address(0)`:

`(block.timestamp << 160) < (expires << 160) | user` holds true for `block.timestamp <= expires`.

And when `user == address(0)`:

`address(uint160((expires << 160) | user))` will give `address(0)` anyway.

So that's how we remove the `iszero`.

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,133 @@
...
+                packed,
+                // `block.timestamp <= expires ? 1 : 0`.
+                lt(
+                    shl(_BITPOS_EXPIRES, timestamp()),
+                    packed
+                )
+            )
+        }
+        return address(uint160(packed));
+    }
```
Vectorized, line 69: ## Explanation

First, we know:

`packed = (expires << 160) | user`

The straightforward way of implementing `block.timestamp <= expires` in is:

`!(block.timestamp > (packed >> 160))` which in Yul, is: `iszero(gt(timestamp(), shr(160, packed)))`

--- 

Now, notice that if `user != address(0)`:

`(block.timestamp << 160) < (expires << 160) | user` holds true for `block.timestamp <= expires`.

And when `user == address(0)`:

`address(uint160((expires << 160) | user))` will give `address(0)` anyway.

So that's how we avoid using an `iszero`.

Vectorized: It is useful to have an optimized `userOf` function for on-chain verification, which is likely a potential use case.

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        if (mayNeedClearing) {
+            // If either `user` or `expires` are non-zero.
+            if (_packedUserInfo[startTokenId] != 0) {
+                delete _packedUserInfo[startTokenId];
+                emit UpdateUser(startTokenId, address(0), 0);
+            }
+        }
+    }
```
Vectorized, line 132: Since this function is called upon mint, transfer, and burn, it is very important to optimize it heavily.

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,126 @@
...
```
Vectorized, line 125: This hook is called upon mint, transfer, burn.

As such it is important to heavily optimize it.

---

The expression:

`!(from == address(0) || from == to)` 

is equivalent to:  

`from != address(0) && from != to`.

`quantity == 1` always holds true whenever `from != address(0)`, so we don't need to check for that.

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,126 @@
...
+        assembly {
+            // Branchless `packed *= block.timestamp <= expires ? 1 : 0`.
+            packed := mul(
+                packed,
+                // `block.timestamp <= expires ? 1 : 0`.
+                lt(shl(_BITPOS_EXPIRES, timestamp()), packed)
+            )
+        }
+        return address(uint160(packed));
+    }
```
Vectorized, line 66: ## Explanation

First, we know that:

`packed = (expires << 160) | user`

The straightforward way of implementing `block.timestamp <= expires` is:

`!(block.timestamp > (packed >> 160))` which in Yul, is: `iszero(gt(timestamp(), shr(160, packed)))`

--- 

Now, notice that if `user != address(0)`:

`(block.timestamp << 160) < ((expires << 160) | user)` is true for `block.timestamp <= expires`.

And when `user == address(0)`:

`address(uint160((expires << 160) | user))` evaluates to `address(0)` anyway.

So that's how we avoid the extra `iszero` opcode.

---

Having an optimized `userOf` function is important for on-chain verification by third party contracts.

contracts/extensions/ERC4907A.sol
Vectorized, line 125: This hook is called upon mint, transfer, burn.

As such it is important to heavily optimize it.

---

The expression:

`!(from == address(0) || from == to)` 

is equivalent to:  

`from != address(0) && from != to`.

---

`quantity == 1` always holds true whenever `from != address(0)`, so we don't need to check for that.

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,126 @@
...
+                or(
+                    // Whether it is a mint (i.e. `from == address(0)`).
+                    iszero(and(from, _BITMASK_ADDRESS)),
+                    // Whether the owner is unchanged (i.e. `from == to`).
+                    eq(and(from, _BITMASK_ADDRESS), and(to, _BITMASK_ADDRESS))
+                )
+            )
+        }
+
+        if (mayNeedClearing) {
```
0xanders, line 118: // for gas savings,  the following lines of code can be deleted
        // if the new owner want to change the user to zero address ,  the new owner could  call `setUser` 
        if (mayNeedClearing) {
            // If either `user` or `expires` are non-zero.
            if (_packedUserInfo[startTokenId] != 0) {
                delete _packedUserInfo[startTokenId];
                emit UpdateUser(startTokenId, address(0), 0);
            }
        }
    }

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,126 @@
...
+     */
+    function _beforeTokenTransfers(
+        address from,
+        address to,
+        uint256 startTokenId,
+        uint256 quantity
+    ) internal virtual override {
+        super._beforeTokenTransfers(from, to, startTokenId, quantity);
+
+        bool mayNeedClearing;
```
0xanders, line 102: //  for gas savings,  the following lines of code can be deleted
 //  if the new owner want to change the user to zero address ,
 //  the new owner could  call `setUser` 
        bool mayNeedClearing;
        assembly {
            // Branchless `!(from == address(0) || from == to)`.
            // Saves 60+ gas.
            // The addresses are masked with `_BITMASK_ADDRESS` to
            // clear any non-zero excess upper bits.
            mayNeedClearing := iszero(
                or(
                    // Whether it is a mint (i.e. `from == address(0)`).
                    iszero(and(from, _BITMASK_ADDRESS)),
                    // Whether the owner is unchanged (i.e. `from == to`).
                    eq(and(from, _BITMASK_ADDRESS), and(to, _BITMASK_ADDRESS))
                )
            )
        }

        if (mayNeedClearing) {
            // If either `user` or `expires` are non-zero.
            if (_packedUserInfo[startTokenId] != 0) {
                delete _packedUserInfo[startTokenId];
                emit UpdateUser(startTokenId, address(0), 0);
            }
        }

0xanders:
Suggestion: Add an  readme file  erc4907a.md  to the docs folder.

Vectorized:
@0xanders Definitely. Readying a separate PR for that soon. :)

Thanks for your work on the EIP. Simple, practical, elegant.

contracts/extensions/ERC4907A.sol
Vectorized: Ah. It’s just to stay faithful to the original implementation. 

I personally feel that resetting user upon transfer like the original code is a good default behaviour.

Let me think again if it is better to reset or keep the user by default.

Vectorized: @cygaar After some marinating, I'm feeling it's better to remove the auto user reset, as suggested.

Let the users add it on their own if they want.

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,87 @@
...
+ */
+abstract contract ERC4907A is ERC721A, IERC4907A {
+    // The bit position of `expires` in packed user info.
+    uint256 private constant _BITPOS_EXPIRES = 160;
+
+    // Mapping from token ID to user info.
+    //
+    // Bits Layout:
+    // - [0..159]   `user`
+    // - [160..223] `expires`
```
Atarpara, line 25: As per your setUser and userOf logic bits layout is

// - [96..159] `expires`
// - [160..255] `user`

Vectorized: Nope. 

If this is the case, the tests will fail as we use `address(uint160(packed))` to get the address.

Atarpara: Solidity Explicit Conversion

```
uint32 a = 0x12345678;
uint16 b = uint16(a);   // b = 0x5678
```

Example : 
packed = 0x000000000000000062D2F4C571C7656EC7AB88B098DEFB751B7401B5F6D8976F

uint160(packed) = 0x71C7656EC7AB88B098DEFB751B7401B5F6D8976F

Vectorized: We define the bit at index 0 as the least-significant-bit, and the bit at 255 as the most-significant-bit.

contracts/ERC721A.sol
```diff
@@ -613,6 +613,22 @@ contract ERC721A is IERC721A {
...
+     * Requirements:
+     *
+     * - `tokenId` must exist.
+     */
+    function _isApprovedOrOwner(address spender, uint256 tokenId) internal view virtual returns (bool) {
+        address owner = ownerOf(tokenId);
+        bool result = true;
+        if (spender != owner)
+            if (!isApprovedForAll(owner, spender))
+                if (getApproved(tokenId) != spender) result = false;
```
cygaar, line 628: I'm realizing now that we have both `getApproved` and `_getApprovedAddress` which are very similar. I'd like our code to be a bit more cohesive - can we cut this down to one function or at least call _getApprovedAddress in getApproved? Right now this feels fragmented.

contracts/ERC721A.sol
```diff
@@ -613,6 +613,22 @@ contract ERC721A is IERC721A {
...
     }
 
+    /**
+     * @dev Returns whether `spender` is allowed to manage `tokenId`.
+     *
+     * Requirements:
+     *
+     * - `tokenId` must exist.
+     */
+    function _isApprovedOrOwner(address spender, uint256 tokenId) internal view virtual returns (bool) {
```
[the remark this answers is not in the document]
cygaar: Follow up to that, lines 651-655 do the same thing as this function - we're checking if a tokenId can be transferred by the msg sender. I'm almost of the mindset that we should just use this function instead of those lines so the code is easier to read at the cost of some extra gas. Either way, I think we need to reduce duplicate logic.

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * @dev Returns the user address for `tokenId`.
+     * The zero address indicates that there is no user or if the user is expired.
+     */
+    function userOf(uint256 tokenId) public view virtual returns (address) {
+        uint256 packed = _packedUserInfo[tokenId];
+        assembly {
+            // Branchless `packed *= block.timestamp <= expires ? 1 : 0`.
```
cygaar, line 55: This is clever

cygaar: But let's change the comment to "Branchless `packed *= (block.timestamp <= expires ? 1 : 0`)." to be clear on the order of operations

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,87 @@
...
+            packed := mul(
+                packed,
+                // `block.timestamp <= expires ? 1 : 0`.
+                lt(shl(_BITPOS_EXPIRES, timestamp()), packed)
```
cygaar, line 59: Is this correct? Let's say timestamp == userInfo.expires. `packed` will end up being greater than `timestamp << 160` because of the bits that represent the user in the first 160 bits of `packed`. Wouldn't you need to zero out the user bits before making a comparison?

Also your comment says <= but here you're using `lt` and not `lte`

contracts/mocks/ERC4907AMock.sol
```diff
@@ -0,0 +1,23 @@
...
+// Creators: Chiru Labs
+
+pragma solidity ^0.8.4;
+
+import '../extensions/ERC4907A.sol';
+
+contract ERC4907AMock is ERC721A, ERC4907A {
+    constructor(string memory name_, string memory symbol_) ERC721A(name_, symbol_) {}
+
+    function safeMint(address to, uint256 quantity) public {
```
cygaar, line 12: I think we should use mint in our mock just so ppl stop trying to use safeMint when they don't need to

cygaar (changes requested):

contracts/ERC721A.sol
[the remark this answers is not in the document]
Vectorized: I think it is ok to duplicate logic.

The Solidity compiler refuses to inline any function with branching.

contracts/ERC721A.sol
Vectorized: I can rename it to `_getApprovedAddressAndSlot`.

contracts/extensions/ERC4907A.sol
Vectorized: ok

contracts/mocks/ERC4907AMock.sol
Vectorized: ok

contracts/extensions/ERC4907A.sol
Vectorized: There is no `lte` in yul. The equivalent is `iszero(gt(..)`.

There is a hidden bitwise trick here:

- If `timestamp()` and `expires` are equal, the existence of an address in the lower 160 bits will allow the `lt` to evaluate to true too.

- In the absence of an address in the lower 160 bits, then the function will return the zero address regardless of the boolean condition.

I've fuzz tested this in Foundry.

contracts/ERC721A.sol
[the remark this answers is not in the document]
cygaar: I'm talking about readability here - it's becoming more and more confusing to track what function does what.

contracts/ERC721A.sol
Vectorized: `getApproved` is part of the ERC721 spec, and has the logic of ensuring the `tokenId` exists.

`_getApprovedAddress` is meant to return both the slot and the value of an approved address without checks. 

I don't think calling `_getApprovedAddress` in `getApproved` will do much to improve aesthetics,   
since it getting the value is simply a one-liner

```solidity
_tokenApprovals[tokenId].value
```

I'll rename `_getApprovedAddress` to `_getApprovedSlotAndAddress` to signify that the function has some low-level meaning.

contracts/extensions/ERC4907A.sol
cygaar: What if `timestamp()` is slightly greater than `expires`, but the existence of an address in the lower 160 bits causes the less than evaluation to be true?

contracts/ERC721A.sol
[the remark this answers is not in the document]
Vectorized: I think this extension is a good opportunity to reintroduce the `_isApprovedOrOwner` function too. 

Can be convenient for 3rd party extensions.

contracts/extensions/ERC4907A.sol
Vectorized:
```
<-   upper 96 bits  -><-           lower 160 bits            ->
[     timestamp      ][0x00000000000000...00000000000000000000]  `timestamp << 160`
[      expires       ][                user                   ]  `packed`
```

There are several conditions:
- `timestamp > expires`:
    - If `user != 0`, then `(timestamp << 160) > packed`, which causes `packed` to be multiplied by 0.
    - If `user == 0`, then `(timestamp << 160) > packed`, which causes `packed` to be multiplied by 0.
- `timestamp < expires`:
    - If `user != 0`, then `(timestamp << 160) < packed`, which causes `packed` to be multiplied by 1.
    - If `user == 0`, then `address(uint160(packed))` will give 0.
- `timestamp == expires`:
    - If `user != 0`, then `(timestamp << 160) < packed`, which causes `packed` to be multiplied by 1.
    - If `user == 0`, then `address(uint160(packed))` will give 0.
    
 Note that `expires` and `user` won't overlap, since we have done the proper castings in `setUser`, which implicit masks them.

cygaar: Ok yeah this makes sense, ignore my last question. The logic here makes sense although I'm a bit worried about ppl reading this and not quite understanding how this works.

contracts/ERC721A.sol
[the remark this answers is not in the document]
cygaar: I would rename this function to be `_isSpenderApprovedForToken` to make it clearer that you're acting on a tokenId. In addition to that, I don't like the term `spender` in ERC721A because this concept is tied closer to 4907. 

I still don't think we need in 721A because extensions can still call isApprovedForAll and getApproved if they really need this information.

Vectorized:
I think we can remove the `_isApprovedOrOwner`.

It's copied from OpenZeppelin though.

contracts/ERC721A.sol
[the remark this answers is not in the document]
Vectorized: Oh ya... why did I miss this!

Vectorized:
I think I'll undo any changes to `ERC721A.sol` in this PR, and move them over to PR #377 instead.

contracts/extensions/ERC4907A.sol
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * - The caller must own `tokenId` or be an approved operator.
+     */
+    function setUser(
+        uint256 tokenId,
+        address user,
+        uint64 expires
+    ) public virtual {
+        address owner = ownerOf(tokenId);
+        if (_msgSenderERC721A() != owner)
```
cygaar, line 42: Let's add a comment here mentioning the gas savings of not having branching similar to what we have in 721A

cygaar (approved):
One comment but otherwise lgtm. We should add in docs in the future

contracts/extensions/ERC4907A.sol
Vectorized: The if-if-if will branch.

The innards of ERC721A uses assembly to compute `msg.sender == owner || msg.sender == approvedAddress` branchlessly.

cygaar: What's the reason behind just making it 3 if statements instead of a single if?

Vectorized: The following actually costs more gas 

```Solidity
if (_msgSenderERC721A() == owner || 
    isApprovedForAll(owner, _msgSenderERC721A() ||
    getApproved(tokenId) != _msgSenderERC721A())
```

PR #387, opened 2022-07-22
Remove projects.md

cygaar:
We originally had this file to showcase projects that used ERC721A. However we're at a stage where a large percentage of projects have switched to 721A. This list probably encompasses a small portion of 721A projects, and has served its original purpose.


cygaar:
Will merge after I update erc721a.org

PR #388, opened 2022-07-25
Fix "error:0308010C:digital envelope routines::unsupported"

nazariyv:
When I have cloned the repo, I proceeded to use node versions 18 and 17 to run tests. Which both failed due to the SSL breaking change in 17 & 18.

Solution is to use node version 16.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [14.x,17.x,18.x]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


Vectorized:
I prefer adding this chunk of code to `hardhat.config.js`.

Idea gotten from https://stackoverflow.com/a/72219174/3686594.

```js
/**
 * ripemd160 algorithm is not available anymore in NodeJS 17+ (because of lib SSL 3).
 * In that case, silently replace ripemd160 by md5 algorithm.
 */
try {
  crypto.createHash('ripemd160');
} catch (e) {
  console.warn('Crypto "ripemd160" is not supported anymore by this Node version');
  const origCreateHash = crypto.createHash;
  crypto.createHash = (alg, opts) => {
    return origCreateHash(alg === 'ripemd160' ? 'md5' : alg, opts);
  };
}
```

cygaar:
Can you add node 17 or 18 to the github action: https://github.com/chiru-labs/ERC721A/blob/main/.github/workflows/run_tests.yml#L21

cygaar (approved):

PR #394, opened 2022-07-28
packedOwnership as lengthless array

sillytuna:
`_packedOwnership` access was performing a `sha3` op for every new index but as these are consecutive they can be stored as a regular array offset from a base storage location irrespective of the start token id. Only one `sha3` is required instead of one per index used. Solidity effectively uses this technique for its arrays but also stores array length, which we don't need to do.

Optimised `_packedOwnershipOf` accordingly.

**Pros**
Reduces gas usage slightly in common functions but more significantly in bulk transfers (figures below).

**Cons**
22,772 additional deployment cost
Code readability

**New**
```
·····································|·························|·············|·············|·············|···············|··············
|  ERC721AGasReporterMock            ·  transferFrom           ·      44232  ·      85751  ·      54612  ·            4  ·          -  │
·····································|·························|·············|·············|·············|···············|··············
|  ERC721AGasReporterMock            ·  transferTenAsc         ·          -  ·          -  ·     295158  ·            1  ·          -  │
·····································|·························|·············|·············|·············|···············|··············
|  ERC721AGasReporterMock            ·  transferTenAvg         ·          -  ·          -  ·     295651  ·            1  ·          -  │
·····································|·························|·············|·············|·············|···············|··············
|  ERC721AGasReporterMock            ·  transferTenDesc        ·          -  ·          -  ·     299608  ·            1  ·          -  │
·····································|·························|·············|·············|·············|···············|··············

```
**Previous**
```
·····································|·························|·············|·············|·············|···············|··············
|  ERC721AGasReporterMock            ·  transferFrom           ·      44338  ·      86037  ·      54763  ·            4  ·          -  │
·····································|·························|·············|·············|·············|···············|··············
|  ERC721AGasReporterMock            ·  transferTenAsc         ·          -  ·          -  ·     297319  ·            1  ·          -  │
·····································|·························|·············|·············|·············|···············|··············
|  ERC721AGasReporterMock            ·  transferTenAvg         ·          -  ·          -  ·     298166  ·            1  ·          -  │
·····································|·························|·············|·············|·············|···············|··············
|  ERC721AGasReporterMock            ·  transferTenDesc        ·          -  ·          -  ·     303983  ·            1  ·          -  │
·····································|·························|·············|·············|·············|···············|··············
```

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -39,6 +39,11 @@ contract ERC721A is IERC721A {
         address value;
     }
 
+    // Reference type for packed ownership.
+    struct PackedOwnershipRef {
+        uint256 value;
+    }
+
     // =============================================================
     //                           CONSTANTS
     // =============================================================
@@ -90,6 +95,11 @@ contract ERC721A is IERC721A {
     bytes32 private constant _TRANSFER_EVENT_SIGNATURE =
         0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef;
 
+    // Storage location for packedOwnerships.
+    // Equivalent to keccak256('com.erc721a.packedownership').
+    uint256 internal constant _PACKED_OWNERSHIP_STORAGE =
+        0x14d1d17051a21856adfcd8caa5ecfacc7e7d5e95e199d4bad2e93a5271be8b2d;
+
     // =============================================================
     //                            STORAGE
     // =============================================================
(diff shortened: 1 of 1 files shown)


contracts/ERC721A.sol
```diff
@@ -90,6 +95,11 @@ contract ERC721A is IERC721A {
     bytes32 private constant _TRANSFER_EVENT_SIGNATURE =
         0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef;
 
+    // Storage location for packedOwnerships.
+    // Equivalent to keccak256('com.erc721a.packedownership').
+    uint256 internal constant _PACKED_OWNERSHIP_STORAGE =
```
sillytuna, line 100: May want to change this to private if the concept is accepted.

Vectorized:
O.O 

This will require a major version bump, cuz of the upgradeable repo.

sillytuna:
Understand about the storage slot situ.

I've got another optimisation I need to experiment with which may make this less important (tho it'll still save gas). Will check tomorrow.  It would also affect storage slots though.

PR #400, opened 2022-08-04
Update minor comment in ERC721A.sol

KamaDeFi:
Fix comment in _getApprovedSlotAndAddress that describes the assembly code.

It is equivalent to `approvedAddress = _tokenApprovals[tokenId].value` and not `approvedAddress = _tokenApprovals[tokenId]`


Vectorized:
Nice catch! Thanks!

PR #405, opened 2022-08-07
Fix _toString comment typo

KamaDeFi:
Question: why does the memory pointer have to be 32-byte aligned? I ran the code reserving exactly 78+32 bytes `str := add(mload(0x40), 0x6C)`, and the execution of the function cost 3 less gas.


Vectorized:
It must be aligned, or else other code that assumes a 32 byte alignment will break.

PR #407, opened 2022-08-11
caller or token owner should be able to approve self as operator for …

jrkosinski:
…own tokens. 

**Motivation:** 
As discussed, the EIP doesn't explicitly state that someone can't be their own operator for their own tokens. It has no effect per se, other than clearing any existing operator for the same tokens. An address being its own approver does no harm, and does not seem to be explicitly forbidden by the EIP. 

approve(my_address, my_token_id) -> results in getApproved(my_token_id) returning the address of the owner. Now the token owner has all the same rights that they already had anyway. 

Same for setApprovalForAll(my_address) -> the token owner is now the operator for all of its own tokens. 

**Changes**
- removed check for (operator == sender) in setApprovalForAll. 
- removed error ApproveToCaller (no longer being used) 
- modified unit test that was testing for that error, to check instead that the call is not reverted. 
- and then modified that same test to check that isApprovedForAll returns true for the token owner 
- added a new test in "setApprovalForAll" to check that similarly, an address can call approve to set itself as operator for its own tokens 
- added a further test that, once an address has called approve to set itself as operator, transferring will clear the approval (as it would for any other operator)

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -459,8 +459,6 @@ contract ERC721A is IERC721A {
      * Emits an {ApprovalForAll} event.
      */
     function setApprovalForAll(address operator, bool approved) public virtual override {
-        if (operator == _msgSenderERC721A()) revert ApproveToCaller();
-
         _operatorApprovals[_msgSenderERC721A()][operator] = approved;
         emit ApprovalForAll(_msgSenderERC721A(), operator, approved);
     }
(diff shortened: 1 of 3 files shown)


Vectorized (approved):
:)

Vectorized:
@jrkosinski Psst, can you help fix the linter issues for the test file?

jrkosinski:
Hi Vectorized,

Yeah, I saw that; I will fix those issues tonight (in a few hours in my
timezone)

Thanks,
John

Vectorized:
@cygaar LGTM, what do you think?

jrkosinski:
👍

cygaar (approved):

Issue #408, opened 2022-08-15
Soulbound Extension


skarard:
The  [Soulbound NFT](https://vitalik.ca/general/2022/01/26/soulbound.html) is burnable but non-transferable and becomes part of the owner's web3 identity. 

The address in the `to` argument of the `_mint(to, quantity)` function will receive a Soulbound NFT. 
The extension will inherit from the ERC721ABurnable extension as the token owner will always be able to burn their token. 

The implementation is to modify `_beforeTokenTransfers(from, to, startTokenId, quantity)` to only allow transfer to and from the null address.

I believe enabling this simple extension for Soulbound NFTs in this repo will pave the way for web3 identity forming NFTs in the future. 

This design of a Soulbound NFT is an extension I will be utilizing in my project, I would like to share my code upstream.

skarard:
I have opened a PR with an example implementation: https://github.com/chiru-labs/ERC721A/pull/410

PR #410, opened 2022-08-15
The Soulbound NFT is burnable but non-transferable and becomes part of the owner's web3 identity

skarard:
This PR is for the most basic implementation of a burnable & non-transferable [Soulbound NFT](https://vitalik.ca/general/2022/01/26/soulbound.html).

This implementation is to modify `_beforeTokenTransfers(from, to, startTokenId, quantity)` to only allow transfer to and from the null address.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,16 @@ _We highly recommend reading the migration guide_, **especially** _the part on [
 
 ## About The Project
 
+### Soulbound NFTs
+This fork of [ERC721A](https://github.com/chiru-labs/ERC721A) is an implementation of an SBT (Soulbound NFT) or a non-transferable NFT (NTT). This token is soulbound to the owner of the token at mint. The token owner is able to burn their token at will.
+
+Soulbound NFTs are an ideal methodology for creating a web3 identity. A community member's web3 identity will be curated from token present on their wallet. 
+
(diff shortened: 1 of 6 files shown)


cygaar:
Appreciate the PR, but this is more of an implementation than an extension.

I'll leave this open for a day or two in case someone wants to see the implementation.

Vectorized:
I feel that this PR is better left for the devs to implement due to the highly open-ended nature of Soul Bound Tokens. 

The devs may want to soulbound:
- only for a certain time period. 
- after a locked flag has been set.
- only certain tokenIds.
- selectively based on some other fancy on-chain interaction logic.

It is difficult to predict what is the exact use case, and we don't want to be opinionated on that here.

Additionally, it is orthogonal to the main goal of ERC721A, which is to provide efficient batch mints.

skarard:
Thanks for the response and for keeping this PR open for a little bit in case anyone is interested. 

As I mentioned in https://github.com/chiru-labs/ERC721A/issues/408 this implementation is to be used in an upcoming project.

I'll keep the Soulbound NFT available for anyone who's interested in SBTs or non-transferable NFTs in my repo https://github.com/skarard/ERC721A on the main branch.

PR #364, opened 2022-07-10
Change tokenApproval slot calculation to avoid compiler error

jameswenzel:
This _should_ mostly be a nit/cleanup thing, but I was running into errors when compiling with `--via-ir` in [a project I'm working on](https://github.com/jameswenzel/bound-layerable).

Some combination of factors led the IR pipeline to fail, presumably when trying to put the slot pointer onto the stack outside of the assembly block:

```
❯ forge build
[⠆] Compiling...
[⠊] Compiling 25 files with 0.8.15
[⠒] Solc 0.8.15 finished in 1.61s
Error: 
Compiler run failed
InternalCompilerError: Invalid stack item name: slot
```

Referencing the slot directly in the `mstore` seemed to fix it on my end, and shouldn't cost any extra gas.

PR #412, opened 2022-08-23
Clarify comment on TokenApprovalRef

Vectorized:
To prevent people from going wtf.

Full explanation: https://github.com/chiru-labs/ERC721A/pull/364


cygaar (approved):
Probably worth leaving a link to where the bug is discussed

PR #413, opened 2022-08-24
Change startTokenId from 0 to 1.

SahalAmeen:
You have to understand. Nobody, and I mean, Nobody starts their tokenIDs from 0. All the smart contracts I coded required me to override the function. Each. and. every. time.

Please just make it start from 1. Like, 90% of the projects out there start with tokenID 1.


Vectorized:
This is a breaking change. 

We deliberately set the default to 0 so that devs can value-add for their non-programmer clients by changing it back to 1. 

Won’t merge. 

Sorry for the inconvenience. Thank you for your understanding.

PR #426, opened 2022-09-24
create an internal _approve function allowing extensions and EIPs to do approvals without check

dievardump:
The current implementation of ERC721A have `_tokenApprovals` private and has all the approval management code in the public facing `approve` function.

This sadly means that there is no way, for a contract extending ERC721A, to internally do approvals without the ownership or isApprovedForAll check.

_There are however some EIPs meant to improve ERC721 and tokens management that are in need of being able to approve internally without those checks._

An example is [EIP-4494](https://eips.ethereum.org/EIPS/eip-4494) that introduces Permits to ERC721, allowing to give approvals to accounts by only signing a message instead of having it done on-chain. Very similar to EIP-2612 for ERC20. 
This EIP could totally change the way we work with marketplaces today, as we would be able to simply sign messages in order to sale items, instead of granting approvalForAll to contracts.

However, EIP-4494 requires the ability to set the approval internally, or at least to bypass the ownership/approvalForAll check, which is right now not possible with the current implementation.

Even if not to accommodate other EIPs and extensions, there can be a lot of reasons for children contracts to be able to do that, similar to the reasons for contracts to be able to `burn` items internally without that check. 

**So, as it was already done with the `_burn()` function, this pull request internalize the content of the `_approve` function with a version that has a 3rd parameter, to indicate if the ownership & isApprovalForAll needs to be checked or not before setting the approval.**

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -409,29 +409,14 @@ contract ERC721A is IERC721A {
     // =============================================================
 
     /**
-     * @dev Gives permission to `to` to transfer `tokenId` token to another account.
-     * The approval is cleared when the token is transferred.
-     *
-     * Only a single account can be approved at a time, so approving the
-     * zero address clears previous approvals.
+     * @dev Gives permission to `to` to transfer `tokenId` token to another account. See {ERC721A-_approve}.
      *
      * Requirements:
      *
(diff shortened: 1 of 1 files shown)


dievardump:
Sorry, didn't see I was on main, redoing.

PR #431, opened 2022-10-13
Optimize ERC721AQueryable for smaller bytecode size

Arkay92:


Vectorized:
Right now, I think Queryable should be optimized for smaller bytecode size. Runtime gas is second priority. 

https://github.com/Vectorized/ERC721A/blob/bytecodeopt/contracts/extensions/ERC721AQueryable.sol

Vectorized:
**Before:**

```
·-----------------------------------|-------------|---------------·
|  Deployments                      ·             ·  % of limit   │
····································|·············|················
|  ERC721AQueryableMock             ·    1460247  ·        4.9 %  │
····································|·············|················
|  ERC721AQueryableStartTokenIdMock ·    1528641  ·        5.1 %  │
·-----------------------------------|-------------|---------------·
```

**After:**

```
·-----------------------------------|-------------|---------------·
|  Deployments                      ·             ·  % of limit   │
····································|·············|················
|  ERC721AQueryableMock             ·    1447033  ·        4.8 %  │
····································|·············|················
|  ERC721AQueryableStartTokenIdMock ·    1517582  ·        5.1 %  │
·-----------------------------------|-------------|---------------·
```

Vectorized:
@cygaar 

After the changes, the deployment gas is 1356357.

This is about 100k gas less, which is around 500 bytes saved. 
Before, the Queryable extension alone adds 2000 bytes. With these changes, it adds 1500 bytes.

Can be useful for people trying to avoid hitting the max contract size.

contracts/extensions/ERC721AQueryable.sol
```diff
@@ -89,56 +111,76 @@ abstract contract ERC721AQueryable is ERC721A, IERC721AQueryable {
...
-            uint256 tokenIdsIdx;
-            uint256 stopLimit = _nextTokenId();
             // Set `start = max(start, _startTokenId())`.
             if (start < _startTokenId()) {
                 start = _startTokenId();
             }
+            uint256 stopLimit = _nextTokenId();
             // Set `stop = min(stop, stopLimit)`.
-            if (stop > stopLimit) {
+            if (stop >= stopLimit) {
```
Vectorized, line 120: Tbh, this `>=` is so that the compiler doesn't add an `iszero` behind the scenes. Functionally equivalent to `>`.

cygaar (approved):

PR #441, opened 2022-12-21
gas optimizations ✨

codeislight1:
a more efficient approach to save on runtime reverts (30 gas) + deployment cost (7500 gas per revert message) is to revert using assembly and explicitly assign bytes4 error message hash, Win-Win. Also, there are cases where SLOAD is being performed before a revert, which is better to be assigned after, so on reverts doesn't go to waste. a contract was deployed after those changes and the deployment  gas cost saving is 82053.

--Code--

### DIFF
--- a/contracts/ERC721A.sol
+++ b/contracts/ERC721A.sol
@@ -465,8 +465,15 @@ contract ERC721A is IERC721A {
      * Emits an {ApprovalForAll} event.
      */
     function setApprovalForAll(address operator, bool approved) public virtual override {
-        _operatorApprovals[_msgSenderERC721A()][operator] = approved;
-        emit ApprovalForAll(_msgSenderERC721A(), operator, approved);
+        address owner = _msgSenderERC721A();
+        _operatorApprovals[owner][operator] = approved;
+
+        assembly{
+            // Emit the `ApprovalForAll` event.
+            let ptr := mload(0x40)
+            mstore(ptr,approved)
+            log3(ptr, 0x20, _APPROVAL_FOR_ALL_EVENT_SIGNATURE, owner, operator)
+        }
     }
 
     /**
@@ -602,7 +609,11 @@ contract ERC721A is IERC721A {
             }
         }
 
-        emit Transfer(from, to, tokenId);
+        // Emit the `Transfer` event.
+        assembly{
+            log4(0, 0, _TRANSFER_EVENT_SIGNATURE, from, to, tokenId)
+        }
+
         _afterTokenTransfers(from, to, tokenId, 1);
     }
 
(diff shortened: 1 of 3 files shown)


cygaar:
Can you run the gas reporter to see the diff between this version and the old version?

Vectorized:
Don't merge first... my OCD.

contracts/ERC721A.sol
```diff
@@ -809,11 +859,22 @@ contract ERC721A is IERC721A {
      * Emits a {ConsecutiveTransfer} event.
      */
     function _mintERC2309(address to, uint256 quantity) internal virtual {
-        uint256 startTokenId = _currentIndex;
-        if (to == address(0)) revert MintToZeroAddress();
-        if (quantity == 0) revert MintZeroQuantity();
-        if (quantity > _MAX_MINT_ERC2309_QUANTITY_LIMIT) revert MintERC2309QuantityExceedsLimit();
+        assembly {
+            if eq(to, 0) {
```
Vectorized, line 863: not safe. Upper bits of to may not be clean.

Vectorized:
Try this first:
```
    function _revert(bytes4 errorSelector) internal pure {
        assembly {
            mstore(0x00, shr(224, errorSelector))
            revert(0x1c, 0x04)
        }
    }
 ```

Vectorized:
It's ok if the revert is after the SLOAD if it is for readability. Cuz no one will submit a tx that will incur the revert on purpose.

contracts/ERC721A.sol
codeislight1: can you elaborate a bit, since when 'to' is passed as an address, it should be a 160 bits value. and the remaining bits supposedly should be set to 0 value.

codeislight1:
> Try this first:
> 
> ```
>     function _revert(bytes4 errorSelector) internal pure {
>         assembly {
>             mstore(0x00, errorSelector)
>             revert(0x00, 0x04)
>         }
>     }
> ```

i have compared it and found that explicitly defining the error is cheaper than having it in a _revert function, but i believe if it's for readability purpose than it's definitely better.

codeislight1:
> errorSelector

yeah that's true, it's more about finding the right balance. but in that case, the  startTokenId is not being used in the if statements checks, so i thought better to come after it. I will revert that change.

contracts/ERC721A.sol
Vectorized: The upper 96 bits may not be cleaned, depending where `to` comes from.

When using regular Solidity, the compiler automatically cleans it lazily when writing to memory, storage, or doing comparisons. No guarantees in assembly.

codeislight1: ohhh i see, thanks for the clarification :)

Vectorized:
I've edited your main branch.

codeislight1:
🫡

codeislight1:
with the optimizer on with 200 runs, the deployment gas cost saving is 26102 gas.

Vectorized (approved):

PR #444, opened 2023-01-18
Add `ERC721ABatchBurnable` extension

jjranalli:
This PR introduces the `ERC721ABatchBurnable` extension, leveraging the `_batchBurn` function introduced here and then refactored in #450.

## Assumptions
Some assumptions had to be made:
- Batch operations accept a uint256[] `tokenIds` param, instead of `startTokenId` + `quantity`. This was done to:
  - Allow the function to be used the same way as other common batch burns
  - Allow to operate on non-consecutive tokenIds, thus being less limiting and more practical for the caller
- The `tokenIds` argument assumes ordered ids
  - It is delegated to the caller making sure that tokenIds are in ascending order.
  
## Notes
- Added minor gas optimizations in `_batchBurn`
- Added tests for `ERC721BatchBurnable`


Vectorized:
O.O Sort.

Vectorized:
Oh, possible to add in an `approvalCheck` parameter to the function? 

Similar to `_burn(tokenId, approvalCheck)`.

In case someone has a use case for doing batch transfers internally.

jjranalli:
> Oh, possible to add in an `approvalCheck` parameter to the function?

Absolutely, on it. Just to confirm, it would result in the following functions:
- `_batchTransferFrom(from, to, tokenIds)`
- `_batchTransferFrom(from, to, tokenIds, approvalCheck)`
- `_safeBatchTransferFrom(from, to, tokenIds)`
- `_safeBatchTransferFrom(from, to, tokenIds, approvalCheck)`
- `_safeBatchTransferFrom(from, to, tokenIds, _data)`
- `_safeBatchTransferFrom(from, to, tokenIds, _data, approvalCheck)`

Vectorized:
@jjranalli Yes!

jjranalli:
@Vectorized approvalCheck flag added

contracts/ERC721A.sol
```diff
@@ -645,10 +645,235 @@ contract ERC721A is IERC721A {
...
+                // - `address` to the next owner.
+                // - `startTimestamp` to the timestamp of transfering.
+                // - `burned` to `false`.
+                // - `nextInitialized` is left unchanged.
+                _packedOwnerships[startTokenId] = _packOwnershipData(
+                    to,
+                    // TODO: Change to this? Costs +115 gas per [i] loop but ensures correctness of NEXT_INITIALIZED in startTokenId.
+                    // If not consecutive set `nextInitialized` to true, otherwise keep it as is.
+                    // (tokenIds[i + 1] != startTokenId + 1 ? _BITMASK_NEXT_INITIALIZED : prevOwnershipPacked & _BITMASK_NEXT_INITIALIZED) |
+                    (prevOwnershipPacked & _BITMASK_NEXT_INITIALIZED) | _nextExtraData(from, to, prevOwnershipPacked)
```
jjranalli, line 728: @Vectorized Logic should now be fixed and slightly more optimized. There was an issue with `nextInitialized` that could've resulted in it not being set correctly.

I have just a question. Here (and in line 783) we can actually derive the correct `nextInitialized`, but combining them it would come with performance tradeoffs that scale with the number of tokens transferred (can be very roughly approximated to an increase of 115 gas per token transferred). I'm currently looking into how logic can be refactored to decrease the impact of this. 

But I'm wondering, would it be fine to just leave `nextInitialized` unchanged, like it is now? That would be the most performant outcome, with main diff being that the `nextTokenId` slot could be initialized while the flag of `startTokenId` or `nextTokenId - 1` is not updated.

jjranalli:
The commit above is an attempt to ensure correctness of `nextInitialized` for all token IDs, while minimizing gas impact.

contracts/ERC721A.sol
Vectorized: It’s ok if the batch transfers don’t touch  `nextInitialized`. Whatever leads to minimal code.

Vectorized:
I think the sort can use insertion sort. Then we can add a comment on passing in sorted tokenIds for best performance. 
The full intro sort costs like one extra kilobyte of bytecode. 


Thanks so much for the PR. 

I’m wondering if we can also do batch burning.

jjranalli:
Perfect, makes sense! Switched to insertion sort.

I had to bump prettier-plugin-solidity as it was [causing an error during lint](https://github.com/chiru-labs/ERC721A/actions/runs/3975760610/jobs/6869639247) preventing checks to go through. However that also caused formatting to change across other contracts. Lmk how you want me to handle that

Will take care of tests now

> I’m wondering if we can also do batch burning
Absolutely, we can use same concept on burn too. Happy to take care of that as well, maybe in a separate PR

jjranalli:
@Vectorized Apologies for the long wait! This is now ready for review

### Notes
- [Fixed issue](https://github.com/chiru-labs/ERC721A/pull/444/commits/7643010d2d0cb6638848ff0e975d26aaabd214a6) in `_batchTransferFrom` that could result in the tokenId subsequent to the last element in `tokenIds` not being initialized correctly
- Written tests for batch transfers and approvals

jjranalli:
> I had to bump `prettier-plugin-solidity` as it was [causing an error during lint](https://github.com/chiru-labs/ERC721A/actions/runs/3975760610/jobs/6869639247) preventing checks to go through. However that also caused formatting to change across other contracts.

Also restored formatting to how it was, so formatting is unchanged. Hopefully checks will go through now.

jjranalli:
@Vectorized It's done! 🫡 I ended up adding batch burns directly in this PR as it leverages common logic to that written for batch transfers, hope that's not an issue

### Main changes
**Batch transfers**
- fixed a bug that prevented `lastTokenId + 1` to be initialized depending on params passed
- previously initialized slots had the `nextInitialized` unchanged during batch transfers. Now they are set to `false` and (unless `extraData` is present) sets to 0 any consecutive tokenId in order to maximise gas savings without compromising functionality
- more tests

**batchBurn**
- shares similar logic of `batchTransferFrom`, with some differences related to params passed and intended behaviour
- ERC721ABatchBurnable extension, interfaces and mocks have been added
- tests

### Notes
- I took the liberty to [refactor some existing code](https://github.com/chiru-labs/ERC721A/pull/444/commits/8c3f49751cc448fc2d47545176bf7630606773df) using the added `_updateTokenId` to reduce bytecode size and improve readability.

jjranalli:
@Vectorized should I make changes to this PR based on #450?

jjranalli:
@Vectorized Managed to have a look at #450, awesome stuff over there. A few questions while I merge it in here and clean things up:

- What is the rationale behind the `burner` param instead of using `msg.sender`?
- `_beforeTokenTransfers` and `_afterTokenTransfers` are executed within the inner do-while loop, while storage variables are updated in the parent for loop during mini-batches. This can cause unexpected consequences if the hooks need to access `_packedOwnerships`, `_packedAddressData` or `_burnCounter`. Is this acceptable? If so it needs to be properly documented. Otherwise the logic I proposed accounted for this by using two separate hooks, but I'm aware it can be inconvenient for different reasons.

Nlferu:
Hi @Vectorized 

Are there any updates on `batchTransferFrom` and `batchBurn` extensions? Any chance those will be included into ERC721A soon?

PR #458, opened 2023-03-26
Add `batchTransferFrom` and `ERC721ABatchTransferable` extension

jjranalli:
This PR introduces `_batchTransferFrom` and safe variants, as well as the `ERC721ABatchTransferable` extension. 

The logic was initially introduced in #444, then redesigned based on `_batchBurn` introduced in #450.

## Assumptions
Some assumptions had to be made:
- Batch operations accept a uint256[] `tokenIds` param, instead of `startTokenId` + `quantity`. This was done to:
  - Allow the function to be used the same way as other common batch burns
  - Allow to operate on non-consecutive tokenIds, thus being less limiting and more practical for the caller
- The `tokenIds` argument assumes ordered ids
  - It is delegated to the caller making sure that tokenIds are in ascending order.


jjranalli:
@Vectorized we should be good here

PR #469, opened 2023-07-28
revert before mint operation to save gas

Santiago0412:


Vectorized:
The calls to the target must be after the state changes, as the target may need to read the updated state.

PR #486, opened 2024-08-21
Add `ERC721ABatchTransferable` and `ERC721ABatchBurnable`

Vectorized:
Continued from #458 and #444.

Notes:
- Refactored some stuff to return `uint256` instead of `address`. 
- Did some minor optimizations. Changing `== 0` to `== uint256(0)` helps the compiler produce more efficient bytecode.
- Doing an array access will incur a bounds check. Thus we use assembly to directly read the memory array values.
- Even though the assembly helper methods are private, we'll give them a uncommon name to prevent collision. Private methods can collide in namespace.


contracts/ERC721A.sol
```diff
@@ -712,6 +703,192 @@ contract ERC721A is IERC721A {
...
+
+    /**
+     * @dev Transfers `tokenIds` in batch from `from` to `to`.
+     *
+     * Requirements:
+     *
+     * - `from` cannot be the zero address.
+     * - `to` cannot be the zero address.
+     * - `tokenIds` tokens must be owned by `from`.
+     * - `tokenIds` must be strictly ascending.
```
axefinaxe, line 725: is there a benifit if the Ids are not only ascending, but also sequential? i.e. 0-4, 5-100 etc

axefinaxe: question answered below in code:
if (tokenId != end)
                    if (_packedOwnerships[tokenId] == uint256(0)) _packedOwnerships[tokenId] = prevOwnershipPacked;

chiru-labs (approved):

Niferu:
This is huuuge!!! Thanks a lot @jjranalli @Vectorized @chiru-labs @cygaar amazing job 🔥🔥🔥

